PR #11, opened 2020-09-15
A si colab rebuild

niklundgren:
@ZKC19940412 I didn't need to change anything to build the docs. You may have had an issue if you don't have latex installed on your machine, or if you don't have some of the right python packages in your environment. See the doc_requirements.txt for details.


gbarbalinardo:
Please don't rebuild the docs, since now we have continuous integration in place

PR #23, opened 2021-03-08
Feature/participation ratio

niklundgren:
Add's the capability to calculate the participation ratio by using the Phonon.calculate_participation_ratio() command.


ZKC19940412:
Looking good.  One suggestion (can be done before or after merging to the master), can you add the demo of this new feature in your example? Or you can send me an example script and then I add it in other clathrate examples.

niklundgren:
I was thinking the same, ya I'll add it in on the prb example and maybe
somewhere for the crystal too.

ZKC19940412:
I think just adding in the prb example will be sufficient, as it is an interesting phenomenon to study for amorphous system and I think the usage (I can be wrong) for participation ratio calculations should be the same for amorphous and crystal. If the usage is different, then it is worth to add in for both amorphous and crystal.

PR #40, opened 2021-09-07
Change third order projections to double precision by default.

ZKC19940412:
Proceed as the name suggested.


gbarbalinardo:
Let's wait before merging this after we are sure it's needed.

PR #48, opened 2022-07-14
update cond qhgk: s_ij for crystal is antihermitian with purely imag …

afiorent:
Solved issue about crystal thermal conductivity with qhgk wrong, often even negative. The generalized group velocity s_ij is an anti hermitian matrix for crystals. The formula should contain |s_ij|^2 not s_ij^2


gbarbalinardo (approved):
LGTM!

PR #55, opened 2022-09-21
  Add logging symmetry when calculating finite difference

dafolkner:
Add logging symmetry when calculating finite difference


gbarbalinardo (approved):
LGTM

PR #61, opened 2022-10-29
colab demo fix by using latest lammps cmake.

ZKC19940412:
Fix broken symlink of ASE-lammps by using the latest lammps cmake.


gbarbalinardo (approved):
LGTM

PR #63, opened 2022-11-01
Kaldo unitcell swapping

ZKC19940412:
Fix the issue of non right-handed convention cell when using lammps externally at the level of generating lammps input files (refer to line 17-21 of 0_generate_LAMMPS_input_and_supercell_structures.py in CNT example for more detail) . Notify the users with documentations in markdowns of the examples using LAMMPS USER-PHONON about this right-handed convention in case they prepare lammps input files from scratch.


gbarbalinardo (approved):
Looks good to me.
Thanks for the detailed comments.

PR #70, opened 2022-11-03
modify plotter to intake user-defined bandpath.

ZKC19940412:
The autogenerated band path from spglib is not always correct so a new flag  is added to allow users pass in the path defined by user manually. This is a feature similar to the one used in ASE Phonons.


gbarbalinardo (approved):
LGTM

PR #86, opened 2023-04-27
Fix bandwith issues.

ZKC19940412:
Change the default to full bandwidth and consistent both qhgk calculations.


gbarbalinardo:
@ZKC19940412 any luck getting the tests to work?

ZKC19940412:
This hidden bug might be deeper than we thought and we need to set up systematic test and fixes for it. Nik and I will spend some time on it after the ACS.

ZKC19940412:
Related issue has been moved to internal doc, so we close it here for now.

PR #89, opened 2023-07-04
Nwl/unfold dispersion redundancy

niklundgren:
1. Improves documentation of Phonon object and usage of the `is_unfolding` flag which previously was read from multiple sources during different routines. We now only accept this argument in the Phonon object.
2. Updates package requirements of users who want to build the docs to be compatible with latest versions on pip
3. Merged in Giuseppe's clean up of the dtypes, with one edit to specify a smaller int type when reading sparse third order matrices to preserve good memory scaling.

Small note on keeping the API understandable-
There should probably be some discussion around whether the "is_unfolding" flag should be moved to the ForceConstants object given that it's more of a treatment of the underlying force constants more than how we treat any phononic  quantity directly.


niklundgren:
This closes #87, and is compatible with zekun's fix for the bandwidth. @ZKC19940412 I'll merge this into main, if you fetch and merge your PR we can close that too!

gbarbalinardo:
Thanks @niklundgren . LGTM!

gbarbalinardo (dismissed):

gbarbalinardo:
Thanks @niklundgren . 
I've noticed some of the tests fail.
```

=========================== short test summary info ============================
FAILED kaldo/tests/test_amorphous_af_conductivity.py::test_af_conductivity_50 - AssertionError: 
Items are not equal to 2 significant digits:
 ACTUAL: 0.002467084722307332
 DESIRED: 0.098
FAILED kaldo/tests/test_amorphous_af_conductivity.py::test_af_conductivity_300 - AssertionError: 
Items are not equal to 2 significant digits:
 ACTUAL: 0.019437854938486515
 DESIRED: 0.532
FAILED kaldo/tests/test_amorphous_classic.py::test_qhgk_conductivity - AssertionError: 
Items are not equal to 2 significant digits:
 ACTUAL: 0.7554788441322309
 DESIRED: 0.996
FAILED kaldo/tests/test_antiresonant.py::test_af_conductivity_without_antiresonant - AssertionError: 
Items are not equal to 3 significant digits:
 ACTUAL: 1.0709869486512578
 DESIRED: 0.804
FAILED kaldo/tests/test_antiresonant.py::test_af_conductivity_with_antiresonant - AssertionError: 
Items are not equal to 3 significant digits:
 ACTUAL: 1.212987825732782
 DESIRED: 0.825
FAILED kaldo/tests/test_antiresonant.py::test_af_conductivity_without_antiresonant_gauss - AssertionError: 
Items are not equal to 3 significant digits:
 ACTUAL: 0.5491826375576702
 DESIRED: 0.8305
FAILED kaldo/tests/test_antiresonant.py::test_af_conductivity_with_antiresonant_gauss - AssertionError: 
Items are not equal to 3 significant digits:
 ACTUAL: 0.5532732461010444
 DESIRED: 0.8335
FAILED kaldo/tests/test_diffusivity.py::test_diffusivity - AssertionError: 
Arrays are not almost equal to 3 decimal
```
Not sure if it's related to this latest change or the change in the bandwidth. But in your [previous commit](https://github.com/nanotheorygroup/kaldo/pull/89/commits/74d4f133c1c38d89cd2bac9b9192ea138e3786bc) they looked ok.

kaldo/observables/harmonic_with_q.py
```diff
@@ -302,24 +298,26 @@ def calculate_eigensystem_unfolded(self, only_eigenvals=False):
...
         supercell_positions = self.second.supercell_positions
         supercell_norms = 1 / 2 * np.linalg.norm(supercell_positions, axis=1) ** 2
-        dyn_s = np.zeros((n_unit_cell, 3, n_unit_cell, 3), dtype=np.complex)
+        dyn_s = np.zeros((n_unit_cell, 3, n_unit_cell, 3), dtype=complex)
         supercell_replicas = self.second.supercell_replicas
         for ind in range(supercell_replicas.shape[0]):
             supercell_replica = supercell_replicas[ind]
-            replica_position = np.tensordot(supercell_replica, cell, (-1, 0))
+            replica_position = np.tensordot(supercell_replica, cell.T, (-1, 0))
             distance = replica_position[None, None, :] + (atoms.positions[:, None, :] - atoms.positions[None, :, :])
```
gbarbalinardo, line 306: Nice catch, can you replace it with 
`replica_position = np.tensordot(supercell_replica, cell, (-1, -1))`
or
`replica_position = np.tensordot(supercell_replica, cell, (-1, 1))`
instead, please? 
This way you don't need to calculate the transpose at every step.

kaldo/observables/harmonic_with_q.py
```diff
@@ -302,24 +298,26 @@ def calculate_eigensystem_unfolded(self, only_eigenvals=False):
...
             projection = (contract('la,ija->ijl', supercell_positions, distance) - supercell_norms[None, None, :])
             mask = (projection <= 1e-6).all(axis=-1)
             neq = (np.abs(projection) <= 1e-6).sum(axis=-1)
             weight = 1.0 / neq
             coefficient = weight * mask
             if coefficient.any():
+                image = supercell_replica%supercell
```
gbarbalinardo, line 313: Can you add a comment on what is `image` or use a more meaningful name, please?
This way we know next time. :)

kaldo/observables/harmonic_with_q.py
```diff
@@ -335,15 +333,15 @@ def calculate_dynmat_derivatives_unfolded(self, direction=None):
...
+        ddyn_s = np.zeros((n_unit_cell, 3, n_unit_cell, 3), dtype=complex)
         fc_s = self.second.dynmat.numpy()
         fc_s = fc_s.reshape((n_unit_cell, 3, supercell[0], supercell[1], supercell[2], n_unit_cell, 3))
         supercell_positions = self.second.supercell_positions
         supercell_norms = 1 / 2 * np.linalg.norm(supercell_positions, axis=1) ** 2
         supercell_replicas = self.second.supercell_replicas
         for ind in range(supercell_replicas.shape[0]):
             supercell_replica = supercell_replicas[ind]
-            replica_position = np.tensordot(supercell_replica, cell, (-1, 0))
+            replica_position = np.tensordot(supercell_replica, cell.T, (-1, 0))
```
gbarbalinardo, line 344: Same as before, please use
`np.tensordot(supercell_replica, cell, (-1, -1))`

gbarbalinardo (approved):
I've left a few minor comments. Good job. 
After addressing these minor changes, please go ahead and merge the branch.

niklundgren:
I removed some small things that I want to change in a future PR because I am not sure they were tested.
Specifically-
1. coords_to_write
file: kaldo/interfaces/eskm.py    line: 107
Our current approach uses python built-in "int" type which is 64 bit to describe coordinates of the third order force constant matrix. When using amorphous systems this easily becomes 100's of Mb if not Gb.
I'd like to switch to using np.uint16 which should cut the memory requirements by 75%

2. is_amorphous
file: kaldo/observables/harmonic_with_q.py    line: 138
Our current approach to detecting whether a system is amorphous is fragile, as it is taken as arguments in some functions, but then explicitly determined in some routines based on replication.
We should calculate this once in the forceconstants/phonons object, possibly allow users to override our decision (e.g. if they want to use a crystal but replicate in reciprocal space only) and pass this internally as needed rather than determine at every step.

3. zheev
file: kaldo/observables/harmonic_with_q.py    line: 327
At the end of calculate_eigensystem_unfolded function, we change our return based on whether or not the user only wanted eigenvalues however it doesn't speed up the calculation at all if they did.
A small if/else statement and one argument passed to zheev could speed up calculations where only the frequency is needed (e.g. dispersion).

PR #95, opened 2023-10-27
Shengbte d3q

afiorent:
Modification of the shengbte interface to read the 3rd forceconstant file from d3q. Now there are 3 shengbte formats: "shengbte", "shengbte-qe" and ""shengbte-d3q".
 Below there is a test on crystalline Germanium between d3q linewidths and kaldo linewidth computed using the interface. The subtle differences can likely be attributed to the different smearing methods in computing third_bandwidth.
Previous and obsolete branch read_d3q can be canceled now.
Please test it and also let me know if you think it is necessary to add an example.

image (1)


afiorent:
I have seen there are some np.float to be converted  in float in the commits, I am postponing the merge

PR #109, opened 2024-02-26
Fix broken install from lfs.

ZKC19940412:
Find a solution for the broken installation due to lfs （original post: https://cloud.tencent.com/developer/ask/sof/106698951） and tested it works for kALDo installation at colab.


gbarbalinardo (approved):
👍

PR #113, opened 2024-03-11
Docker containers that will allow the GPU/CPU to function for kaldo.

rebeltroopers:
Docker files that work for the GPU on kaldo. Zekun has tested it and it works for all future things that KALDo wants. Deepmd and lammps are also included in the build.

Let me know what you want me to add as well.


Docker-files/Dockerfile-dev
```diff
@@ -0,0 +1,55 @@
+# Use Ubuntu 22.04 as the base image
+FROM ghcr.io/deepmodeling/deepmd-kit:devel_cu11
```
gbarbalinardo, line 2: @rebeltroopers what is `devel_cu11`? I can't understand it from the[repo](https://github.com/deepmodeling/deepmd-kit/pkgs/container/deepmd-kit/190590505?tag=devel_cu11).

Docker-files/README.md
```diff
@@ -0,0 +1,78 @@
+# KALDodocker
```
gbarbalinardo, line 1: Cool name! Let's just call it `Docker files for kALDo`.

Docker-files/LICENSE
```diff
@@ -0,0 +1,674 @@
+                    GNU GENERAL PUBLIC LICENSE
```
gbarbalinardo, line 1: No need to add the license if it's a subfolder of kALDo.

Docker-files/README.md
```diff
@@ -0,0 +1,78 @@
...
+
+## Overview
+KALDodocker is a Docker container designed for computational simulations, integrating LAMMPS with Python support and KALDo software. This container is tailored for CPU environments but also includes instructions for adapting it for GPU usage.
```
gbarbalinardo, line 4: It should be clear what the difference between the three versions is:
- only kaldo
- kaldo + deepmd
- kaldo + deepmd + cuda

Docker-files/requirements.txt
```diff
@@ -0,0 +1,17 @@
+numpy==1.25.1
```
gbarbalinardo, line 1: Why is there a different `requirements.txt` and not the same one that kaldo uses?

gbarbalinardo (commented):
Very nice work @rebeltroopers . I added a few comments.
The dockerfiles for DeepMD can be very helpful. Please update the readme in the example with DeepMD to reference this folder.

Docker-files/Dockerfile-dev
rebeltroopers: devel_cu11 uses CUDA11 as a base image. The devel is for development version, as you need it to compile with nvcc. So it would be deepmd using cuda 11 with all development tools as well.
https://hub.docker.com/r/nvidia/cuda

Docker-files/README.md
rebeltroopers: Ok. I will rename them. the goal was as follows.

- kalldo-dev is gpu and cpu enabled but also install the git repository for testing. This would be for developers for an environment.
- Kaldo-CPU was for kaldo with cpu only usage. It only really has kaldo and nothing else.
- Kaldo-GPU includes deepmd and other software. Zekun had helped me discuss items that would be implemented in the future, so it has some additional items.

I will go rename the files to be clearer.

Docker-files/README.md
rebeltroopers: Will do.

Docker-files/requirements.txt
rebeltroopers: Good point. I will update the requirments.txt to be kaldo's. I will go resolve any dependency issues and tell you if that is the case.

Docker-files/README.md
gbarbalinardo: Got it. We can keep it as it is, if it makes sense. Let's just link it from the DeepMD example.

rebeltroopers:
Docker files have been update with your requests and now only uses the kaldo requirments.txt. Also, I have cleaned up the README.md

gbarbalinardo (approved):
Looks good to me. Please rename the folder to `Dockerfiles` before merging.

rebeltroopers:
Final commit with changes that were asked.

PR #131, opened 2024-07-09
Update doctools.js

### LINKED ISSUE #130: Fix search in documentation
Search broken on documentation, no results are returned and no error message is seen on the page.

niklundgren:
This is a temporary solution to the broken search function. 

Search functionality is restored, however I don't know for certain whether it is a completely safe fix. No unit tests broken., but further testing required.

The .noConflict() that is called is apparently a way of restoring any variables that were "saved" when jquery starts running. see here- https://api.jquery.com/jQuery.noConflict/
However, in this instance it overwrites the value of the _ function jquery uses, which then breaks the search.
It doesn't seem that we need to restore any variables from "previously loaded javascript frameworks" (phrase from docs on noconflict) so this shouldn't be an issue, however it's unclear why it was added in the first place.
Closes #130

### DIFF
--- a/docs/_static/doctools.js
+++ b/docs/_static/doctools.js
@@ -12,7 +12,8 @@
 /**
  * select a different prefix for underscore
  */
-$u = _.noConflict();
+/* $u = _.noConflict(); Experimentally commenting out this line to fix search function. This seems to set the underscore to a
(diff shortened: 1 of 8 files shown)


gbarbalinardo (approved):
Nice job!

PR #133, opened 2024-07-17
Added another safety check to pdos routine

Sassafrass6:
Allow p_atoms to be passed to pdos routine as None, in which case the total phonon dos is calculated. Previously, this safety check was only done in kaldo/controllers/plotter.py, so there was no default behavior for unspecified p_atoms, if phonons.pdos was called directly by the user.


gbarbalinardo (approved):
LGTM!

PR #135, opened 2024-08-09
Sigma2 TDEP in forconstants

rebeltroopers:
Sigma2 for a TDEP only is added to the forceconstants class. I am not as familiar with class structure so let me know if it needs to be redon.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -6,8 +6,8 @@
 .. image:: docsource/_resources/logo.png
    :width: 400
 
-A Software to Predict Thermal Properties Using Lattice Dynamics.
-=======================================================================
+kALDo: Advanced Thermal Property Predictions via Lattice Dynamics and Machine Learning for Nanoscale Materials.
+===============================================================================================================
 
 Read the paper on Arxiv_.
 


niklundgren:
commits all look fine, just need to pull main first.

ZKC19940412 (commented):
Tested the functionality at the moment, and encountered some issues (i.e. cyclic inputs, undefined variables). Need to fix them before merge. Preferable with a test case to run pytest on the other end. Also need to test on more materials. Tried it with my current explorations of BaTiO3, BaZrO3 and BaHfO3. It seems the sigmaA derived from here is lower than I would expect. Will send in some benchmarks using phonopy/fhi-vibes soon.

kaldo/forceconstants.py
```diff
@@ -273,3 +277,35 @@ def elastic_prop(self):
         
         # Denote parameter for irreducible Cij in the unit of GPa
         return evperang3togpa * cijkl / evtotenjovermol
+
+
+    def sigma2_td_md(self,fc_file='infile.forceconstant', primitive_file='infile.ucposcar', supercell_file='infile.ssposcar', md_run='dump.xyz'):
+        initial_structure = read(sc_file, format="vasp")
+        second_order_fc_in_full_dimension     =   parse_tdep_forceconstant(
```
gbarbalinardo-tonal, line 284: Code formatting is not correct.

gbarbalinardo, line 284: Please update the code formatting to follow [Python standards](https://peps.python.org/pep-0008/). One space is enough.

PR #147, opened 2024-10-13
Split third order calculation to run on a single nu (phonon index) at the time. 

gbarbalinardo:
14 Squashed commits:
[c680c4a] Revert changes on `storage.py`
[5720c70] Split third order calculation to run on a single nu (phonon index) at the time. 
[290b1b8] Change order of parameters in single nu calculations 
[c196ff0] Add documentation
[0b70086] Fix error in parameters names
[d255d0f] Code cleanup
[7b6e7cc] fix typo
[7305077] Minor fix
[8ce725f] Reverted units update
[783c270] Add single mu for amorphous
[53cf056] Refactor single mu method for crystal
[15f2ba7] Update storage helper
[8687eaa] Clean up decorator for lazy loading.

PR #158, opened 2025-01-17
Third order projection progress tracker

niklundgren:
Motivation:
In a recent commit pulled to main, the logging information was removed from the third order projections. However, these projections make up the bulk of CPU hours used when running heat transport simulations with kALDo. Therefore, logging the progress of these projections allows users to benchmark performance and predict CPU requirements for a given calculation.

Solution:
I reimplemented the progress logging in its original form in two commits (in project_crystal and project_amorphous) I then added two commits that improved the readability of the loggers by using formatted string literals which also reduces the number of function calls.
    - function call removed: np.round()

I also added logic to only output this information every 200 modes in the amorphous (consistent with the frequency of the old logger in the project_crystal function) at the cost of one boolean evaluation per mode, but saving two logging calls on the 199 other passes. 

  
Commit that removed the logger SHA: 925eda35501c36290a544d18f02b9b78e663c4da [link](https://github.com/nanotheorygroup/kaldo/commit/925eda35501c36290a544d18f02b9b78e663c4da)
Merged in PR #147


ZKC19940412:
Tested and it works as expected! Thanks for adding back this feature, it is very useful to gauge the time spent.

ZKC19940412 (commented):
LGTM

ZKC19940412 (approved):

PR #159, opened 2025-01-24
Fix loading VASP force constants file

bkhli:
- Fix loading with `ForceConstants.from_folder` (`format=shengbte`) for VASP force constants
- Add a test case for reading VASP force constants file to calculate thermal conductivity. 
- Clean code and more documentations for elastic constant tests and shengbte_io.py


kaldo/observables/secondorder.py
```diff
@@ -365,9 +365,9 @@ def load(cls, folder, supercell=(1, 1, 1), format="numpy", is_acoustic_sum=False
                     filename = folder + "/FORCE_CONSTANTS"
                 if not os.path.isfile(filename):
                     raise FileNotFoundError(f"File {filename} not found.")
-                second_order = shengbte_io.read_second_order_matrix(folder, supercell)
+                second_order = shengbte_io.read_second_order_matrix(filename, supercell)
                 second_order = second_order.reshape((n_unit_atoms, 3, n_replicas, n_unit_atoms, 3))
-                grid_type = "C"
+                grid_type = "F"
```
gbarbalinardo, line 370: @bkhli . This is a critical line. Have we tested it with different examples?

bkhli: I have tested with Si and Ge example data, but I am not sure how to show they in the test cases, because they are large examples. I can provide the results in the comments though.

gbarbalinardo (approved):
👍

kaldo/observables/secondorder.py
gbarbalinardo: Thanks, approved!

PR #164, opened 2025-02-12
Parallelize pytest to reduce CI time

bkhli:
I tested in my local environment that `pytest-xdis` library can distribute tasks across multiple cores if available. It worth to try this to reduce CI time.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,7 +13,7 @@ jobs:
     steps:
       - checkout
       - run:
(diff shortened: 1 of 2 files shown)


bkhli:
I will merge now and test it out to see if it works.

bkhli:
@gbarbalinardo I need your help. This breaks CI now. This modifies the docker image, which is not updated. How to update the docker image?

PR #170, opened 2025-03-13
I started over to make things easier for the commit

rebeltroopers:
Redid the commit as I think the last one makes no sense from a commit standpoint. It seems to be confusing a bunch of other peoples work. 

Here is the files reedited to make sense as I just started over.

Examples of Kaldo with ase with dftb+.


gbarbalinardo:
LGTM!

gbarbalinardo (approved):
LGTM

PR #171, opened 2025-03-13
I added the q-points to the documentation as a hint. 

rebeltroopers:
I also fixed a typo that was breaking the math that was above it.


gbarbalinardo (approved):
Nice work.

PR #173, opened 2025-03-22
Fix network issue when download_isotope_data in a better way

bkhli:
I fixed this bug by locating the path to the library and find the json file. 

This fixes requires python verison >= 3.9.


gbarbalinardo:
Does it require pip install?

bkhli:
This is part of standard library after python >= 3.9. It does not require pip install a new library.

gbarbalinardo (approved):
Thanks! LGTM.

PR #177, opened 2025-04-30
Initial NAC correction push

niklundgren:
Changes here:
Initial functions to handle BEC and dielectric tensor provided they are in QE format. They require I'm 

1. because it's not too memory intensive, the velocity correction ideally should be calculated for all three cartesian directions at once. Currently, it's calculated one at a time (because that's how we handle velocities without correction).
2. It should be removed from the "unfolding" routines, as it's been decided this correction is valid for classical potentials which do not need unfolded force constants.
3. Examples to be updated


niklundgren:
I'm aware @bkhli is waiting on this and would like it before Wednesday, so to that end I'm going to bypass the review for now and get it through so we can make some progress here while there's time!

Bohan and I worked in person together for the merge commit to ensure this was thoroughly reviewed.

Note: Reviewing the commit one more time, I'd like to note that I removed the changes of commit [397ebe2](https://github.com/nanotheorygroup/kaldo/pull/177/commits/397ebe2107b20d980b11336026227f6a8efb6fee) because it should be separate from the feature add here.

PR #157, opened 2025-01-16
Fix for wrong bulk modulus with dynmat read from QE and various bug fixes and cleanups

bkhli:
I made a temporary fix for wrong bulk modulus with dynmat read from QE (`format == shengbte`) in `kaldo/forceconstants.py` and created a unit test for it. 

List of bug fixes and changes: 

- `is_qe_input` should not be always true in `secondorder.py:load`
- `is_acoustic_sum` is shorted in `secondorder.py:from_supercell`. 
- various undefined functions and variables in `secondoder.py:remap_force_constants`
- Add documentation and code readability for `elastic_prop` and `read_second_order_qe_matrix`

PR #179, opened 2025-05-25
Properly fix elastic property

bkhli:
This commit properly fixes wrong elastic property due to force constant format.

We can add some test later when I am more free. 

Replace commit 3697134 in #157.


gbarbalinardo (approved):
Looks good to me.

PR #183, opened 2025-07-01
Gb/free energy calculator

gbarbalinardo:


dafolkner (approved):
Methods are consistent and are within error of PBE ice paper.

PR #185, opened 2025-07-03
Zc/quick fix on docs

ZKC19940412:
Use another format to render citations in documentations. Manually merged free energy implementation to this branch.


niklundgren:
Thanks for the updates @ZKC19940412

PR #186, opened 2025-08-03
CLI tool and integrate with kALDO logging

gbarbalinardo:
Integrate command line tool for `kaldo`
  - Propagate verbose flag to kALDO's internal logger
  - Plotter config to map directly to API functions
  - CLI documentation


niklundgren (approved):
Nice, should probably set some unit tests up for it at some point

PR #189, opened 2025-08-10
Multiple temperatures speedup

gbarbalinardo:
Multi-T speedup: avoids rerunning heavy Dirac-delta and third-order contractions.

We decoupled temperature-invariant terms from temperature-dependent populations:
- Cache `sparse_phase` and `sparse_potential`.
- At new temperatures, recompute only population factors and contract.
- `project_amorphous` calls +/- channels separately.
- `calculate_ps_and_gamma` consumes cached tensors +/- and populations.


niklundgren (approved):
This looks really great - awesome feature. Again hard to catch things by eye here at 13 files changed but I went through it.

As a matter of record - I primarily see changes to imports, and small changes around adding the new "Storable" parent class, in addition to adding new _store_formats properties under each class.
Two new unit tests are a nice add-on too.

It may be nice to display maybe some of the features brought here, as well as, the new CLI tools in an example specifically oriented around the practical side of data I/O and the user interfaces (python + cli)

niklundgren:
Since this likely conflicts with my PR #191 (both editing harmonic_with_q.py) and I'm waiting on review - I'll pull this into main and recommit my branch.

PR #193, opened 2025-08-15
Here is a working example with the new dependancies. So far the docs …

gbarbalinardo:
…look fine compiled. Should be around python3 3.11 and with numpy 2.1


rebeltroopers:
Fails to show latex docs. I missed it when I was tired last time. I will need to go figure out what is going on. I think that section may need to be redone to if we want compliance with new sphinx. 

I will look at it later in the week. If someone else wants to help you can build it and see the warnings. I think some titles need to be renamed.

PR #196, opened 2025-08-30
Small improvement on calculating forceconstants, compatibility with other libraries, and docs with mathjax

bkhli:
When using a force field to calculate forceconstants, kaldo uses the same calculator twice. To prevent potential issues on some force fields that might have memory leak or data race issue, I removed the unused one. In addition, I changed the syntax to avoid the warning being printed by ASE. 

Right now, kaldo requires the fixed versions of numpy, scipy, ase and pandas. It is easy to cause a package version conflicts with other libraries. For example, mattersim (v1.0-1.1) requires a higher version of ase. It is actually not necessary to fix versions of these libraries, except requiring numpy < 2.0. My proposal here is to use `>=` to replace `==` in requirements.txt. It passed all the tests when installing kaldo with this new requirement.txt file. 

Docs now are rendered with mathjax. Fixed various code block syntax issues over all the doctrings and docs.


gbarbalinardo (approved):
This is awesome @bkhli !

PR #197, opened 2025-09-09
Fix docker image and copyright

bkhli:
Fix the Dockerfile that builds the image that generates the documentation website.


gbarbalinardo (approved):
Looks great!

PR #198, opened 2025-09-14
Fix doc fonts download in CI by using system default fonts

bkhli:
This quick commit fixes the google fonts by replacing with the system default font.


gbarbalinardo (approved):
Looks good!

PR #209, opened 2025-11-03
QHA fix implemented

dafolkner:
This fixes the QHA issue and changes the fit method to 4th order polynomial ridge regression.


gbarbalinardo:
You can probably leverage some of the files in https://github.com/nanotheorygroup/kaldo/pull/206

PR #213, opened 2025-11-07
Nwl/mgo example refactor

niklundgren:
Simplified and fully functional example. Generates dispersion + dos plots using the kaldo tool. Calculates the TC and prints it nicely, but doesn't plot it yet.

Results for TC w/ NAC: 47.5 W/mK
Results for TC w/o NAC: 53.8 W/mK

The force constants used for these examples were pretty large (9x9x9 second order 5x5x5 third order) so I want to rerun with something smaller before I write the test cases. The compressed FC's cost ~50Mb on git-lfs.

I'll write up a test case for the TC ASAP along with the phonon properties plot.


ZKC19940412:
Try pushing all the files multiple times. When I push files in another repo, the rule seems to be "50 MB" per push, not total. You might be able to push the whole FCs after multiple pushing. Meanwhile, keep all the files in ALD other than _sparse potential and __gamma__and__ps (the whole scattering tensor) file will be sufficient to reproduce necessary results.

examples/mgo_espresso_d3q/2_run_simulation.py
```diff
@@ -0,0 +1,128 @@
...
+# tf.config.set_visible_devices(devices[gpu_index])
+
+import numpy as np
+from kaldo.observables.harmonic_with_q import HarmonicWithQ
+from kaldo.forceconstants import ForceConstants
+from kaldo.phonons import Phonons
+from kaldo.controllers.plotter import plot_dos, plot_dispersion
+from kaldo.conductivity import Conductivity
+
+# ============================================================
```
gbarbalinardo: Minor, no need to ascii graphics `======`.

examples/mgo_espresso_d3q/2_run_simulation.py
```diff
@@ -0,0 +1,128 @@
...
+
+# Simulation Settings
+temperature = 300
+classical_stats = False
+# True/False means Classical/Quantum heat capacity + phonon populations
+
+# Correction-dependent Settings
+# True = using NAC
+# False = without NAC
+nac_folder_dic = {
```
gbarbalinardo: Difficult to read. I'd suggest something in the for loop
`io_folder = 'forces' if is_nac_corrected else 'force_no_charges'`

examples/mgo_espresso_d3q/2_run_simulation.py
```diff
@@ -0,0 +1,128 @@
...
+	True: 'forces',
+	False: 'forces_no_charges'
+}
+
+for is_nac_corrected in [True, False]:
+	io_folder = nac_folder_dic[is_nac_corrected]
+	# ============================================================
```
gbarbalinardo: remove `====` everywhere.

gbarbalinardo (commented):
Thank you for working on this, @niklundgren . I added a few minor comments to address. 
I can't see the bug fix in the main kaldo code, did you not push that?

examples/mgo_espresso_d3q/README.md
```diff
@@ -1,11 +1,12 @@
 # Setup
-Please use the "unpack.sh" bash script to unpack the force constants
+Please use the "0_unpack.sh" bash script to unpack the force constants
```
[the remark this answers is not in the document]
niklundgren: it's removed!

examples/mgo_espresso_d3q/1_unpack.sh
```diff
@@ -4,5 +4,5 @@
 tar -xvzf tools/mgo-qe-d3q.tgz
 for file in POSCAR FORCE_CONSTANTS_3RD
```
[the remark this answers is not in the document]
niklundgren: done!

examples/mgo_espresso_d3q/2_run_simulation.py
niklundgren: ASCII is removed

niklundgren:
this should be up to spec and greatly improved/simplified.
The only thing I feel is maybe too complicated is the plotting script, which I haven't seen accomplished in a less messy way but I'm open to suggestions. I tried to use a minimal amount of variable names and write reusable functions where possible.

PR #214, opened 2025-11-09
Initialize a draft of lammps mix super cell.

ZKC19940412:
While polishing the kALDo example, I also managed to implement mix supercell scheme where forceconstants are imported from LAMMPS.  It will be super useful for universal potential such as MACE or MatterSim. Given the graph neural network nature, universal potential  requires large supercell for 2nd IFCs, such as 10x10x10 and beyond. However, third order is decayed way faster, thus much smaller cell in third order (e.g. 5x5x5) will still ensure convergence in kappa. 

Mix supercell has been supported in ase (native kALDo) and phonopy/phono3py interfaces. My current implementation enables the LAMMPS interface. Unitest is also composed, and tested in local. 

## Summary by CodeRabbit

* **New Features**
  * Added support for loading mixed-supercell LAMMPS force-constant data.
  * Automatically detects the appropriate third-order replicated-atom structure, with fallback support when needed.
  * Provides a clear error when required replicated-atom files are unavailable.

* **Bug Fixes**
  * Corrected third-order data loading for mixed-supercell configurations.

* **Tests**
  * Added regression coverage for loading, fallback behavior, error handling, and conductivity calculations.


gbarbalinardo:
Thanks @ZKC19940412. I like the idea, though I’m a bit concerned it could cause confusion for users. Is there an easy way to add a warning when the sources differ?

ZKC19940412:
What do you mean by source differ?

bkhli (approved):
This feature is useful for crystal systems in lammps. The I/O logic for lammps here is slightly from other formats. For other formats, the size of supercell is inferred from the forceconstant files. Here, it directly uses the input file that feeds into lammps. This choice is sort of natural since one needs a different structure input file to produce force constant in lammps. I think this feature is legit, but it needs a good documentation or example to show its usage before releasing this feature.

gbarbalinardo:
Rebased onto current main. cc @ZKC19940412.

gbarbalinardo (approved):
LGTM.

ZKC19940412:
Thanks. @gbarbalinardo please feel free to merge it when you think ready.

PR #220, opened 2026-01-20
Fix temperature-dependent storage paths in HarmonicWithQTemp

gbarbalinardo:
Add _get_folder_path_components() to properly parse <temperature> and <statistics> placeholders in property labels. Without this, temperature- dependent properties (heat_capacity_2d, population, heat_capacity) were cached to paths missing temperature/statistics components, causing data from one temperature to be incorrectly reused for subsequent temperatures in QHGK conductivity calculations.


ZKC19940412 (approved):
Tested and it worked as expected. Thanks for fixing it. LGTM to merge.

PR #221, opened 2026-01-23
Fix formatted storage shape loss for diffusivity and other tensors.

gbarbalinardo:
Fixes #243

Arrays with leading singleton dimensions (e.g. frequency with shape (1, n_modes) for single-k-point systems) lose that dimension through the np.savetxt/np.loadtxt round-trip. This causes frequency[0] to return a scalar instead of a 1D array, breaking velocity calculation.

Fix: store the original shape in the file header and restore it on load. Also handles split files for 3D arrays with spatial coordinates (velocity).


gbarbalinardo:

PR #222, opened 2026-01-31
Speedup NAC code

bkhli:
Replaced `np.` with `contract` in `HarmonicWithQ` class to speed up NAC code. 

For my system with ~1500 atoms, `frequency` is speed up from 26 mins to 7 mins.


gbarbalinardo (approved):
Looks good. Have you had a chance to test it?

bkhli:
Yes. First, it passed all the unit tests. Second, I also tested with my system to see that it indeed speeds things up. 
For my system with ~1500 atoms, the time to get the frequency property calculated is speed up from 26 mins to 7 mins.

PR #225, opened 2026-02-25
Rename format vasp to vasp-sheng and qe-vasp to qe-sheng

bkhli:
`vasp` is renamed to `vasp-sheng`. `qe-vasp` is renamed to `qe-sheng`

This is to distinguish the second and third order force constant by inventors. VASP and Quantum Espresso only invented second order force constant file formats. Third order force constant file format (denoted as `sheng` here) is invented by ShengBTE in thirdorder.py. 

All the test files and docs are updated.


gbarbalinardo (commented):
@bkhli looks great. @ZKC19940412 does it break any examples?

PR #228, opened 2026-03-19
Introduce Parallel module. Parallelize 2nd + 3rd order FC calculations.

niklundgren:
Adds support for parallelizing the third-order force constants calculation in the finite-difference calculator. Calculations can be distributed across workers using Python's multiprocessing module, with results optionally flushed to disk periodically to cap peak memory usage at N_flush × 9 × N_atoms × 3.

Interrupted calculations can be resumed without restarting from scratch, as completed displacements are detected and skipped automatically.

Both the parallelized calculation, and resume functionality has been tested on a HPC node with 128 threads, resulting in approximately a ~100x speed up over serial calculations.
A multi-node run on 2x128 threads was also confirmed to execute as expected.

A unit test is added that checks for accuracy down to a relative difference of 1e-7

An example is also being added in nanotheorygroup/kaldo-examples/empirical_potentials, with plots to showcase the identical answer for serial/parallel calculations.


niklundgren:
> @niklundgren @ZKC19940412 I did some edits. Feel free to take a look!

Woah, this is very cool! I never stop learning from being on team kALDo.
The abstraction is really nice, I wouldn't have thought to do that. 

Thanks for the small fixes on documentation (like capitalization and bad variable names). 

I'm not sure if we need unit tests for the k-point parallelization - that is technically untested at the moment, but I'm not confident on whether they're truly necessary.

niklundgren:
So to be clear, we're just waiting on a review by @ZKC19940412 to merge? Let me know if you need any support with that, zekun :smile: 
Would love to get this through!

gbarbalinardo:
Hey @niklundgren, I pushed a few follow ups on top of yours:

Added a validator that fails fast when someone passes a non picklable calc such as `MatterSim`, `MACE`, `ORB`, `CPUNEP` with `n_workers > 1`, with a `functools.partial(...)` hint in the error. Also selects `spawn` when `CUDA` is initialized in the parent to avoid the fork plus CUDA crash, and binds the calc to `replicated_atoms` once in serial mode.

Made `SecondOrder.calculate` and `ThirdOrder.calculate` behave the same. Both accept `calculator=None`, both auto resolve `scratch_dir` in parallel runs only. Fixed a crash where `SecondOrder.calculate(calculator=None)` rebinds `None` before `get_forces()`. Dropped `kaldo/parallel/gpu.py` since it is not imported anywhere and consolidated duplicated tests.

Full suite passes on the remote. Verified byte for byte parity with `main` on the `kaldo-examples` pattern. Existing examples do not need changes.

If this works for you I think we are good to merge 👍 .

@ZKC19940412 feel free to take a look.

gbarbalinardo (approved):
Great work!

niklundgren:
I'm good with this! Ready to merge on my end. I think @ZKC19940412 is the last hold-out, I'd love his sign off.

ZKC19940412 (approved):
LGTM, ready to merge. Great job team!

PR #231, opened 2026-03-25
`k`-points parallelization

gbarbalinardo:
## Summary

Parallelizes the per-k-point crystal projection across workers using the
`get_executor()` abstraction introduced in #228, and extracts the
duplicated dispatch+resume loop into one shared helper.

## What's new

1. **`kaldo.parallel.dispatch_with_resume()`** — single helper that owns
   the dispatch + resume + sentinel-write loop, replacing three
   copy-pasted versions. Workers write their own output files; the
   helper writes the `.done` sentinel only after the worker returns
   successfully (atomicity: no sentinel without a complete result).

2. **Parallel per-k-point crystal projection** — `_project_crystal()`
   extracts per-k-point work into a free `_compute_kpoint_projection()`
   function (numpy in, TF tensors created internally so it's
   spawn/forkserver-safe). Each k-point's modes are computed
   independently in its own worker.

3. **Disk-based projection output with resume** — optional
   `projection_output_dir` parameter on `Phonons` writes per-k-point
   `kpt_NNNNN.npz` files plus `.done` sentinels. Interrupted runs
   resume by skipping completed k-points, matching the FC scratch-file
   pattern.

4. **Memory fix in `_project_crystal`** — the previous draft of this
   PR accumulated all per-k-point results in a dict before optionally
   writing them to disk, defeating `projection_output_dir` for large
   k-meshes. Results are now flushed inside the dispatch loop and
   dropped from memory; assembly loads each k-point back from disk one
   at a time.

5. **Migrations** — `calculate_second()`, `calculate_third()`, and
   `_project_crystal()` all now use `dispatch_with_resume()`.
   `displacement.py` shrinks by ~40 lines net.

6. **Spawn-when-TF-loaded** — `_default_mp_context()` already picked
   `spawn` when CUDA was available; it now also does so when
   TensorFlow is imported in the parent process. TF eagerly creates a
   thread pool at import; forking duplicates the FDs but not the
   threads, leaving the worker's TF runtime in a half-initialized
   state. Symptoms range from silent deadlocks to
   `CUDA_ERROR_INVALID_HANDLE`. xdist + parallel kaldo on Linux
   reliably tickles this — exposed by this PR's CI hang on the first
   push.

## What's parallelized now

| Stage | Work unit | Pattern |
|-------|-----------|---------|
| FC generation (#228) | per-atom | `dispatch_with_resume()` + scratch files |
| Crystal projection (this PR) | per-k-point | `dispatch_with_resume()` + output files |
| Amorphous projection | serial (1 k-point) | unchanged |

## Verification

- New unit tests for the dispatcher (`test_parallel_dispatcher.py`):
  empty-input, resume-skips-existing-sentinel, sentinel-on-success,
  no-sentinel-without-output-dir, no-sentinel-on-worker-failure,
  parallel-matches-serial.
- New unit test for the spawn-when-TF behavior in
  `test_parallel_executor.py`.
- New regression test in `test_parallel_projection.py` for the memory
  fix: with `projection_output_dir` set, every k-point's `.npz` exists
  on disk after the run.
- Pre-existing `test_parallel_second.py` and `test_parallel_third.py`
  continue to pass after the dispatcher migration.
- CI green (164 tests).

## Release Notes

* **New Features**
  * Parallel k-point projection for crystalline systems with configurable worker processes
  * Resume capability for interrupted projections using on-disk checkpoints
  * New `n_workers` and `projection_output_dir` parameters in Phonons class

* **Improvements**
  * Enhanced multiprocessing context selection to safely handle TensorFlow and CUDA libraries

* **Refactor**
  * Unified parallel execution workflow for improved consistency and maintainability

* **Tests**
  * Comprehensive test coverage for parallel dispatcher, executor, and projection workflows

### DIFF
--- a/kaldo/controllers/displacement.py
+++ b/kaldo/controllers/displacement.py
@@ -4,7 +4,9 @@ import os
 import numpy as np
 from concurrent.futures import as_completed
 from kaldo.helpers.logger import get_logger
-from kaldo.parallel import get_executor, is_parallel, validate_parallel_calculator
+from kaldo.parallel import (
+    dispatch_with_resume, get_executor, is_parallel, validate_parallel_calculator,
+)
 from sparse import COO
 logging = get_logger()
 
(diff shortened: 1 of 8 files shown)


niklundgren (approved):
I'm not sure if pinning the BLAS/OMP threads to 1 is optimal code practice (I ended up removing it in my branch), might be worth removing in `phonon.py`

The main changes seem to be adding helper functions before the Phonon class that support the new k-point parallelization in `project_crystal` function

Along with the new parallel module for kaldo that will be where the parallelization handling lives.

Overall, a solid PR and extremely exciting new feature

gbarbalinardo:
@niklundgren Good catch — removed in 252e8aa.

Those `os.environ['OMP_NUM_THREADS'] = '1'` assignments were a no-op anyway: libgomp / MKL / OpenBLAS read those env vars at first initialization, which happens during `import tensorflow` / `import numpy` long before `_compute_kpoint_projection` runs. Setting them inside the worker function changes nothing for the already-loaded threadpools.

Worker thread caps are correctly applied by `kaldo.parallel.executor._init_worker_thread_caps`, which runs as the `ProcessPoolExecutor` initializer *before* the worker imports anything, and respects user-set caps via `setdefault`.

Thanks for the review!

PR #234, opened 2026-04-06
Fix dtype mismatch when loading float32 third-order force constants

### LINKED ISSUE #235: QHGK crashes on amorphous systems with float32 third-order force constants
`_project_amorphous` crashes when `third.npz` contains float32 data:

```
InvalidArgumentError: cannot compute SparseTensorDenseMatMul as input #3
was expected to be a float tensor but is a double tensor
```

`ThirdOrder.load` does not enforce a dtype. Float32 third-order data (used by large systems to save memory) mismatches the float64 eigenvectors in `tf.sparse.sparse_dense_matmul`.

Reproduced on nanthy with 1920-atom amorphous carbon. Both `main` and `v2.0.1` are affected.

Fix: #234

gbarbalinardo:
Large amorphous systems (e.g. 1920-atom amorphous carbon) often save `third.npz` as float32 to reduce memory. This causes a dtype mismatch at `phonons.py:1095` where TF requires matching dtypes for sparse-dense matmul. The existing crystal path (`_project_crystal`) already casts to a common dtype, but the amorphous path did not.

This PR, ensure float64 dtype when loading third-order force constants from `third.npz` in `ThirdOrder.load`. Fixes TensorFlow `SparseTensorDenseMatMul` crash in `_project_amorphous` when third-order data is float32 but eigenvectors are float64

Closes #235

### DIFF
--- a/kaldo/observables/thirdorder.py
+++ b/kaldo/observables/thirdorder.py
@@ -92,7 +92,8 @@ class ThirdOrder(ForceConstant):
                               pbc=[1, 1, 1])
 
                 _third_order = COO.from_scipy_sparse(load_npz(os.path.join(folder, THIRD_ORDER_FILE_SPARSE))) \
-                    .reshape((n_unit_atoms * 3, n_replicas * n_unit_atoms * 3, n_replicas * n_unit_atoms * 3))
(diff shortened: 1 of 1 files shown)


niklundgren:
I ended up using the same fix, and it worked - tested on nanthy with the last branch.

PR #237, opened 2026-04-06
Move ballistic early return before scattering matrix computation

### LINKED ISSUE #224: `finite_length_method='ballistic'` unnecessarily computes full scattering matrix
### Description

When using `finite_length_method='ballistic'` with `method='inverse'`, the code in `calculate_mfp_inverse()` computes the full scattering matrix and its inverse (lines 455–469 of `conductivity.py`) before the ballistic block (lines 473–480) overwrites `lambd` with the purely geometric result:

```python
gamma_inv[nonzero_velocity] = length[alpha] / (2 * np.abs(velocity_alpha[nonzero_velocity]))
lambd[physical_mode, alpha] = np.diag(gamma_inv).dot(velocity_alpha)
```

The ballistic mean free path depends only on group velocities (IFC2) and sample dimensions — it does not require IFC3, phonon bandwidths, or the scattering matrix. However, the current code flow triggers these expensive computations (via `self.calculate_scattering_matrix()` and `phonons.bandwidth`) and then discards the results.

### Suggested fix

Add an early return at the top of `calculate_mfp_inverse()` when `finite_length_method == 'ballistic'`:

```python
def calculate_mfp_inverse(self):
    length = self.length
    phonons = self.phonons
    finite_length_method = self.finite_length_method
    physical_mode = phonons.physical_mode.reshape(phonons.n_phonons)
    velocity = phonons.velocity.real.reshape((phonons.n_phonons, 3))
    lambd = np.zeros_like(velocity, dtype=np.float32)

    if finite_length_method == 'ballistic':
        for alpha in range(3):
            if (length[alpha] is not None) and (length[alpha] != 0):
                velocity_alpha = velocity[physical_mode, alpha]
                gamma_inv = np.zeros_like(velocity_alpha)
                nonzero_velocity = velocity_alpha != 0
                gamma_inv[nonzero_velocity] = length[alpha] / (2 * np.abs(velocity_alpha[nonzero_velocity]))
                lambd[physical_mode, alpha] = np.diag(gamma_inv).dot(velocity_alpha)
        return lambd

    # ... rest of the existing code for 'ms' and default inverse
```

This avoids computing IFC3 projections, scattering rates, and the full matrix inversion, making the ballistic limit a genuinely lightweight IFC2-only calculation.

### Impact

- Removes unnecessary computational cost (IFC3 projection + matrix inversion) when only the ballistic limit is needed
- Enables ballistic conductivity calculations for systems where IFC3 are unavailable or too expensive to compute

gbarbalinardo:
When `finite_length_method='ballistic'`, the full scattering matrix was unnecessarily computed for each spatial direction before being discarded. Move the ballistic branch to return early, avoiding the expensive scattering matrix calculation entirely.

Closes #224


gbarbalinardo:

PR #239, opened 2026-04-06
Fix tests for eigenvector basis dependence in degenerate subspaces

### LINKED ISSUE #238: Tests fail on arm64 due to eigenvector basis dependence in degenerate subspaces
## Problem

`test_iso_bw` and `test_qhgk_conductivity` fail on arm64 (Apple Silicon) while passing on x86 CI.

## Root cause

LAPACK's eigensolver returns an arbitrary orthonormal basis within each degenerate eigenspace. The specific basis chosen depends on the platform's LAPACK implementation (Accelerate on arm64 vs MKL/OpenBLAS on x86). For crystals with degenerate phonon modes, any observable that depends on individual eigenvectors within a degenerate group — rather than the subspace projector — will differ across architectures.

Affected quantities:
- **Isotopic scattering rates** (`isotopic_bandwidth`): Tamura's formula involves `|e_{ks} · e_{k's'}|^2` terms that are basis-dependent within degenerate multiplets, even though the trace over the degenerate group is invariant.
- **QHGK conductivity with per-mode bandwidth**: The `diffusivity_bandwidth` defaults to `isotopic_bandwidth`, propagating the basis dependence into the conductivity tensor.

## Fix

- `test_iso_bw`: Assert the mean over the degenerate triplet (modes 3-5) instead of a single mode value
- `test_qhgk_conductivity`: Use a constant `diffusivity_bandwidth=1.0` to test the QHGK formula independently of the per-mode bandwidth basis dependence

gbarbalinardo:
`test_iso_bw` and `test_qhgk_conductivity` fail on arm64 while passing on x86 CI because LAPACK returns different eigenvector bases within degenerate subspaces depending on the platform.

- Fix `test_iso_bw` by asserting the mean of the degenerate triplet (modes 3-5) instead of a single mode value
- Fix `test_qhgk_conductivity` by using `diffusivity_bandwidth=1.0` instead of the default per-mode bandwidth

Closes #238


gbarbalinardo:
Related: #237 (ballistic early return fix, split from the same original work)

gbarbalinardo:

PR #241, opened 2026-04-06
Remove unnecessary gc.collect() calls from conductivity

gbarbalinardo:
Python's reference counting already frees numpy arrays immediately when the last reference is dropped. The explicit gc.collect() calls in conductivity.py added overhead by triggering full generational sweeps without benefit. The `del` statements are kept as they reduce peak memory during the scattering matrix loop.


gbarbalinardo:

PR #245, opened 2026-04-07
Addition of realspace forceconstant symmetry to calculation of IFCs on top of parallelization.

dafolkner:
This is a modified version of the nwl/parallel_third_calculator branch with the realspace symmetry added to reduce number of IFC computations. One method added to perform the symmetry analysis in displacements.py, symmetry flag added to '.calculate()' method for secondorder and thirdorder, IFC calculation methods in displacement.py modified to use reduced atom list derived from symmetry when turned on.


niklundgren (approved):
I tested it on the aluminum EMT example in kaldo-examples and saw an 11x speed-up.

Relevant code (`displacement.py` lines 12:297) is well-commented. 

Not sure about the overall architecture (as to where the symmetry finding function is best located) but this seems fine for now.

This should close #246 when merged.

ZKC19940412:
Does n_workers>4 work for you? If so, can you help Higo resolve that issue? If not, is the 11x speed up achieved from a broken n_workers schema?

niklundgren:
## Regarding @higo02 's observation about 4 workers

ProcessPoolExecutor only spawns as many workers as the number of futures submitted.
The number of futures submitted = len(atoms_to_compute), which equals the number of atoms in the unit cell.

If your unit cell has only 4 atoms, only 4 subprocesses are ever launched — regardless of what you pass to n_workers.

Hence why my system with 1920 atoms can see an 100x speed up, but the smaller unit cells are only 4x speed-up.

## Regarding the Sym benchmark

I was running the symmetry speed-up without any parallelization on the aluminum example, so the 11x speed-up was purely the difference between serial+symmetry vs. serial without symmetry

PR #249, opened 2026-04-14
Symmetry added to parallel q branch

dafolkner:
This PR adds q-point symmetry to the parallel q-point branch developed and pushed by Giuseppe. Flag added to Phonons object to turn symmetry on/off. Turning it on leads kALDo to construct an irreducible q-point list that it loops over instead and then constructs the full tensors accordingly. Applies ONLY to third order properites (i.e. projections), second order properties are left as is, but may be implemented for greater efficiency (though much less significant than savings on projections).

## Summary by CodeRabbit

* **New Features**
  * Added optional q-space symmetry support for phonon calculations (configurable).

* **Improvements**
  * Symmetry-aware computation and caching reduce redundant work and speed up evaluations when enabled.
  * Phase-space and tensor results are propagated from symmetry representatives to all k-points.

* **Tests**
  * Added a test asserting zero workers is rejected (n_workers must be >= 1).
  * Removed a disk-output regression test for per-k-point file writes.


gbarbalinardo (approved):
Added a few small changes. It looks good to me.

gbarbalinardo:
Two CodeRabbit findings addressed in 035aaa5 + 90dd4cc:

**Time-reversal fallback (035aaa5)** — generalized to all rotations. The previous fallback only checked `q_target == -q_irr` (R = I followed by time reversal). For non-centrosymmetric crystals where -I isn't in the point group, equivalence may require composing a non-trivial rotation with time reversal: `q' = -R_recip · q`. The new code iterates rotations and sets `krot_perm = -R.T`, `krot_cart = -A·R·A⁻¹` on match. Si (centrosymmetric, our test case) doesn't reach the fallback so existing behavior is unchanged; the fix lets the symmetry feature work on non-centrosymmetric crystals where the previous code would have raised `RuntimeError`.

**Cache scoping (90dd4cc)** — added a `<use_q_symmetry>` token to the storage label of every cached anharmonic artifact (`_sparse_phase_and_potential`, `_ps_gamma_and_gamma_tensor`, `bandwidth`, `phase_space`, `anharmonic_bandwidth`), mirroring the `<include_isotopes>` pattern. When `use_q_symmetry=True`, artifacts go to `.../qsym/<artifact>`; existing no-sym caches stay at the same paths.

The collision CodeRabbit flagged was real: `_sparse_phase_and_potential` under symmetry has `None` entries for non-IBZ k-points (filled in later by the replication step in `_select_algorithm_for_phase_space_and_gamma`). A subsequent `use_q_symmetry=False` run would load this, find Nones, and skip the replication path (which only fires when `use_q_symmetry=True`), leaving ~85% of k-points with zero scattering — silently wrong conductivity. Cache scoping prevents that.

**Considered alternative:** replicate non-IBZ entries during `_project_crystal` so `_sparse_phase_and_potential` is fully populated and the replication step in `_select_algorithm_for_phase_space_and_gamma` disappears. That would let the cache be shared across modes — but it'd require moving the symmetry replication out of your `_select_algorithm` and into the projection assembly. Bigger algorithmic change than the rebase wanted to take on. The cache-path fix is the smaller intervention; the algorithmic refactor (with shared cache) could be a follow-up if you want it.

PR #252, opened 2026-04-27
Finalized Realspace Symmetry

dafolkner:
Updated realspace symmetry added to updated parallel branch as of 4/27/2026.

### DIFF
--- a/docs/docsource/api_forceconstants.rst
+++ b/docs/docsource/api_forceconstants.rst
@@ -40,18 +40,30 @@ of frames is :math:`2 \times 2 \times 3N \times 3N = (6N)^2`.
 Should I use kALDo to calculate my IFCs?
 ========================================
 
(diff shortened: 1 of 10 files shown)


niklundgren:
same question as other PR about `use_symmetry` + `scratch_dir` behavior -

I noticed in the old branch that the init function for `third_order` objects was setup to remove user input for the `scratch_dir` if they set up a `scratch_dir=<folder>` and also set `use_symmetry=True` - I think this should be swapped to fail instead (just raise an error with an explanation that you can't use both for now) and should that also be duplicated for the second order? (which also can take `scratch_dir` as an argument).

Also, you might try running the sym+scratch calculation because I don't think it should break if only some atoms are calculated - in theory it still works

dafolkner:
Glitched out weirdly.

PR #256, opened 2026-05-03
Trim CodeRabbit tone_instructions to fit 250-char cap

gbarbalinardo:
CodeRabbit rejected the prior config on PR #255 with: "Validation error: String must contain at most 250 character(s) at tone_instructions". The substantive numerical-review guidance (float ==, dtype, einsum, TF top-level imports) lives in path_instructions, which has no such cap, so the tone field can stay short and focus on style of feedback. New length is 196 chars.

## Summary by CodeRabbit

* **Chores**
  * Updated internal code review configuration to enhance consistency and focus on technical accuracy in the development process.

### DIFF
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -4,10 +4,9 @@
 language: en-US
 early_access: false
 tone_instructions: >
-  Be concise and technical. This is a scientific Python library for thermal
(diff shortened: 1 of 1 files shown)


gbarbalinardo:
Folded into #255.

PR #269, opened 2026-06-14
Fix dos by replacing np.traz to spicy.integrate.trapezoid for numpy > 2.4.

ZKC19940412:
This PR addresses the usage of np.trapz for dumpy > 2.4. It breaks the current dos routine silently (i.e. raising a warning during the kaldo simulation but does not break the whole run), yet no dos/pdos will be generated. Substitute from np.trapz to spicy.integrate.trapezoid works and generates proper dos/pdos.

## Summary by CodeRabbit

* **Refactor**
  * Improved the numerical integration used for phonon density of states (PDOS) normalization, switching to a trapezoidal integration approach for more reliable and consistent PDOS scaling.

### DIFF
--- a/kaldo/phonons.py
+++ b/kaldo/phonons.py
@@ -1328,8 +1328,7 @@ class Phonons(Storable):
                 for j in range(proj.shape[1]):
                     p_dos[ip, i] += np.sum(amp * proj[:, j, :] * physical_mode)
 
-            # TODO: np.trapz is deprecated in numpy 2.0+, but np.trapezoid does not exist before numpy 2.0. migrate it after this function gets removed. 
-            p_dos[ip] *= 3 * n_atoms / np.trapz(p_dos[ip], f_grid)
+            p_dos[ip] *= 3 * n_atoms / np.trapezoid(p_dos[ip], f_grid)
 
         return f_grid, p_dos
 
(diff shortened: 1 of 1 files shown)


higo02 (approved):
I think we had this issue before in an older version of kALDo, it was previously fixed and now it came back again.

Just tested this new branch from zekun and the DOS calculation runs okay, good to merge.

kaldo/phonons.py
```diff
@@ -1328,8 +1329,7 @@ def pdos(self, p_atoms=None, direction=None, bandwidth=0.05, n_points=200):
                 for j in range(proj.shape[1]):
                     p_dos[ip, i] += np.sum(amp * proj[:, j, :] * physical_mode)
 
-            # TODO: np.trapz is deprecated in numpy 2.0+, but np.trapezoid does not exist before numpy 2.0. migrate it after this function gets removed. 
-            p_dos[ip] *= 3 * n_atoms / np.trapz(p_dos[ip], f_grid)
+            p_dos[ip] *= 3 * n_atoms / trapezoid(p_dos[ip], f_grid)
```
gbarbalinardo, line 1332: Is there any similar method in Numpy?

ZKC19940412: There is np.trapz for old numpy and np.trapezoid for numpy 2.0+. No universal trapz method on numpy works for both old and new numpy.  Meanewhile, spicy.trapeziod works from old to new version.

gbarbalinardo (approved):
Thanks @ZKC19940412 . LGTM!

PR #272, opened 2026-07-06
Fix replica-order mismatch in the QE force-constant load path

gbarbalinardo:
read_second_order_qe_matrix returns replica axes (t1, t2, t3) which SecondOrder.load flattens in C order (t3 fastest), while the replica grid was declared grid_type=F (t1 fastest). Every force-constant block (t1, t2, t3) was therefore paired with replica vector (t3, t2, t1), which is equivalent to evaluating the dispersion at q with q1 and q3 permuted.

For cubic crystals the permutation is a symmetry operation, so the existing MgO/Si/Ge tests were blind to it. For anisotropic cells the dispersion is wrong everywhere off Gamma: wurtzite GaN had in-plane and out-of-plane branches swapped, with errors above 100 cm^-1 and unphysical degeneracy patterns (validated against QE matdyn.x running on the same espresso.ifc2).

Declare the replica grid C for the qe-* formats, in SecondOrder.load and in ThirdOrder.load. The two must change together: second and third order shared the mislabeling, and fixing only one breaks the joint consistency that the conductivity relies on (verified on the Si/Ge qe test suites).

The vasp-* formats are left exactly as on main: their loaders show the same C-data/F-grid pattern, but second and third order are jointly consistent, the cubic references only validate that joint behavior, and no hexagonal VASP-format validation data is in the test suite yet. Fixing that path is a separate follow-up; notes are left at both declaration sites.

Add a GaN wurtzite regression test with frequencies validated against QE matdyn.x on the same force constants (asr=simple): q-points on and off the c-axis plus the Gamma-A transverse degeneracy the ordering bug broke.

Known remaining issue, unchanged by this fix: both interpolation paths still deviate from matdyn by up to ~35 cm^-1 for hexagonal cells near the A point (replica-image assignment / Wigner-Seitz boundary weights); tracked separately.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Quantum ESPRESSO force-constant handling to preserve the expected tensor layout/axis ordering, improving phonon frequency consistency for QE-derived inputs.
  * Maintained consistent behavior for VASP/shengbte-related formats to avoid regressions.

* **Tests**
  * Added a regression test for wurtzite GaN phonon dispersion using Quantum ESPRESSO force constants (`qe-d3q`).
  * Validates frequencies at selected q-points against reference values and checks the expected transverse-acoustic mode degeneracy.


gbarbalinardo:
While working on the read-side fix here, I noticed the **write-side twin** of this same bug in `save_second_order_qe_matrix` (`kaldo/interfaces/shengbte_io.py:430`), which is not touched by this PR but is worth tracking:

```python
file.write('%4d %4d %4d' % (int(l_vec[2]), int(l_vec[1]), int(l_vec[2])))
```

The replica index is written as `(t3, t2, t3)`: the `t1` column is never emitted and `t3` is duplicated. So any `espresso.ifc2` written by kaldo has a corrupted replica index (correct only when `t1 == t3`, e.g. cubic supercells with equal dimensions), the write-side analogue of the read-side ordering mismatch fixed in this PR. It should read `(int(l_vec[0]), int(l_vec[1]), int(l_vec[2]))`.

There is also a likely related C/F-order question: the reader (`read_second_order_qe_matrix`) returns replica axes `(t1, t2, t3)` and callers now flatten them C-order (this PR), so the writer's `list_of_index` ordering should be checked against that same convention when this is fixed. Round-tripping a known `.fc` through save + load would be a good regression test.

Left out of this PR to keep it focused on the read path; opening a tracking issue.

gbarbalinardo:
Tracking issue for the deferred follow-ups (writer bug, VASP path, hex interpolation residual, NAC): #279

PR #278, opened 2026-07-06
Force Constant Type Heirarchy

ejmeitz:
This PR outlines a type hierarchy I think would be beneficial to adopt for code cleanliness, correctness and extensibility long-term. This is just an outline/proposal of the types and how they might be used. I have not fully implemented or checked correctness. This was inspired by some problems I ran in to while trying to implement the cumulant expansion and some bottlenecks I observed with the existing COO format. 

At a high-level I have added two abstract types: `AbstractForceConstantStorageFormat` and `AbstractForceConstantFormat`. 

IFC Storage Formats:
----------------------
The goal of the storage format type is to enable conversion between internal storage formats easily and to provide a stronger contract and more information about `self.value` in `SecondOrder`. I envision two sub-types to begin: `ReplicaFormat` and `BlockSparseFormat`. The `ReplicaFormat` is what kaldo already uses where force constant are stored in a COO tensor with shape like `(n_uc, 3, n_replica, n_uc, 3)`. `BlockSparseFormat` is what TDEP natively uses and naturally embeds the neighbor list and extends to fourth-order better than a COO array. 

These types implement dunder methods like `__getitem__` so they behave exactly like `self.value` does in `SecondOrder` today, but because they are classes they contain extra metadata, have the ability to enforce/check formatting rules and encapsulate methods to convert between the formats easily.

IFC Format Types:
---------------------
The goal of these types was mainly cleanliness. They pull out the implementation inside `SecondOrder.load` into library specific types. Here I have implemented the `TDEP` type. Most of the logic was already in the `interfaces` folder, but adding a class just helps tie things together and provide better abstraction.

### DIFF
--- a/kaldo/interfaces/AbstractIFCFormat.py
+++ b/kaldo/interfaces/AbstractIFCFormat.py
@@ -1,0 +1,107 @@
+from abc import ABC, abstractmethod
+from enum import Enum
+
+class AbstractForceConstantStorageFormat(ABC):
+
+    def __init(self, ifcs, fmt_type, err_msg):
+        try:
+            self.order = fmt_type.order_map[len(ifcs.shape)]
+        except KeyError:
+            raise KeyError(err_msg)
+        
+        self.ifcs = ifc_tensor
+
+
+    # Define dunder methods so this type acts like a raw numpy array
+    #! MIGHT NEED OTHER ONES
+    def __getitem__(self, index):
+        return self.ifcs[index]
+
+    def __setitem__(self, index, value):
+        self.ifcs[index] = value
+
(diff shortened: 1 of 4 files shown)


gbarbalinardo:
Your block-sparse idea already paid off twice. The dense-IFC4 bottleneck you flagged here is what #274 fixed (the TDEP parser now streams straight into COO, 13 GB -> 38 MB peak on the SW fixture), and the same pattern turned out to be the CI OOM in the F1 path on #283, fixed the same way. Let's pick the full type hierarchy back up once #283 lands and we know which format conversions we actually need. Good call on this one.

PR #281, opened 2026-07-10
Relax QHGK conductivity test tolerance

### LINKED ISSUE #270: QHGK conductivity test is sensitive to numerical backend and fails on Linux x86-64
# QHGK conductivity test is sensitive to numerical backend and fails on Linux x86-64

## Description

The `test_qhgk_conductivity` test fails reproducibly on my Linux x86-64 machine, while it passes on macOS ARM64.

The failing assertion is:

```text
FAILED kaldo/tests/test_io.py::test_qhgk_conductivity - AssertionError:
Items are not equal to 3 significant digits:
 ACTUAL: 139.72576904296875
 DESIRED: 141.0
```

The computed value is deterministic across repeated runs.

## Environment

Linux environment:

```text
OS: Arch Linux
Architecture: x86_64
CPU: AMD Ryzen 5 9600X
Python: 3.10.20
NumPy: 2.2.6
SciPy: 1.15.3
TensorFlow: 2.21.0
KALDO commit: de89b75
BLAS/LAPACK: OpenBLAS
```

The result is consistently:

```text
139.72576904296875
```

Disabling oneDNN and forcing single-thread execution did not change the result.

## Cross-platform comparison

Using the same KALDO commit and matching Python package versions where possible, I observed:

| Platform / numerical backend | QHGK conductivity |
| --- | ---: |
| macOS ARM64, Apple Accelerate | `141.06184` |
| macOS ARM64, OpenBLAS  | `140.34895` |
| Linux x86-64, OpenBLAS | `139.72577` |

Each value was reproducible across repeated runs.

Changing the BLAS/LAPACK stack on the same macOS ARM64 machine shifted the result from:

```text
141.06184
```

to:

```text
140.34895
```

This suggests that the end-to-end QHGK result is sensitive to the numerical backend and possibly also to architecture-specific numerical implementations.

## Full test suite

On the Linux machine, the full test suite reports:

```text
1 failed, 171 passed, 60 warnings
```

The only failing test is:

```text
kaldo/tests/test_io.py::test_qhgk_conductivity
```

## Discussion

The current assertion checks equality to three significant digits. The largest deviation observed from the reference value `141.0` is approximately `0.9%`.

I noticed that other numerical tests in the suite also use the same three-significant-digit convention, so I am not sure whether the preferred solution is:

- using a relative tolerance only for this QHGK test;
- updating the expected reference value;
- investigating the backend sensitivity further;
- or keeping the current convention and documenting the supported numerical stack.

Would a local relative-tolerance check be appropriate here, for example:

```python
np.testing.assert_allclose(
    conductivity,
    141.0,
    rtol=2e-2,
    atol=0.0,
)
```

or would you prefer a different approach?

AlessandroSerra:
## Summary

Relax the QHGK conductivity regression test tolerance to account for the cross-backend spread observed with the default adaptive diffusivity bandwidth.

Different BLAS/LAPACK backends may select different eigenvector bases inside degenerate subspaces, resulting in small differences in the computed conductivity. Across macOS ARM64 with Accelerate, macOS ARM64 with OpenBLAS, and Linux x86-64 with OpenBLAS, the measured values ranged from `139.72577` to `141.06184`.

The updated tolerance covers this observed numerical envelope.

Related to #238.
Closes #270.

## Testing

The following tests were run locally to validate the change:

* `pytest kaldo/tests/test_io.py::test_qhgk_conductivity -v`
* `pytest kaldo/tests/test_io.py -v`

## Summary by CodeRabbit

* **Tests**
  * Updated conductivity validation to use a relative-tolerance comparison, allowing results within 2% of the expected value.

### DIFF
--- a/kaldo/tests/test_io.py
+++ b/kaldo/tests/test_io.py
@@ -36,7 +36,7 @@ def phonons():
 def test_qhgk_conductivity(phonons):
     cond = Conductivity(phonons=phonons, method="qhgk", storage="memory").conductivity.sum(axis=0)
     cond = np.abs(np.mean(cond.diagonal()))
-    np.testing.assert_approx_equal(cond, 141, significant=3)
+    np.testing.assert_allclose(cond, 141.0, rtol=2e-2, atol=0.0)
 
 
 def test_inverse_conductivity(phonons):


gbarbalinardo (approved):
Thank you @AlessandroSerra looks good to me.

PR #282, opened 2026-07-10
Free Energy Cumulant Expansion (rebased)

gbarbalinardo:
## Motivation

This is @ejmeitz's free-energy cumulant expansion (#254), rebuilt on current main. It supersedes #254, which had fallen 103 commits behind after we extracted its loader-layer content into #273/#274; rebasing it in place was no longer practical, and pushing to the fork branch is blocked by LFS lock verification on maintainer-edit access. The full original commit history remains available on #254 and on the fork branch for reference.

**Authorship**: the cumulant module commit (`1c201150`) is authored by Ethan Meitz; the physics (sampler, contractors, F1/F2 kernels, estimator, bootstrap) is his work, validated against his published 25^3 Ne values to 1.7e-7 eV/atom. Please merge with a merge commit or rebase-merge, not squash, so that commit keeps his authorship.

## What changed

Three commits:

1. **Add kaldo.cumulant** (Ethan Meitz): the module, its test suite, and the vendored LJ/SW TDEP fixtures.

   F_total = F_H + F_0 + F_1 + F_2, with a supercell-Gamma canonical sampler (quantum or classical), analytic F1/F2, flat-table Taylor contractors for V2/V3/V4, and a bootstrapped constant-correction estimator mirroring the Julia reference.

2. **Fix Phase-5 atom frame, batch LAMMPS single points, V2_tilde, table**: the four fixes that unblock the constant-correction sampling.
   * Phase 5 now runs in one atom frame (ssposcar order). The sampler drew displacements in replica-major order while the contractors index atoms in ssposcar order, so V2/V3/V4 were evaluated on permuted configurations (measured 7.24 eV vs the correct 2.80 eV). Together with the loader fixes in #274 this takes F_0 on the SW fixture from ~3e12 to a physical -1e-5 eV/atom; with a harmonic surrogate standing in for LAMMPS, `<V> - <V2>` = 1e-14 eV end to end.
   * `BatchEnergyEvaluator`: one persistent LAMMPS instance, per config only `scatter_atoms` + `run 0 pre no post no` + `get_thermo('pe')`, replacing the per-config ASE loop (~0.3 s/config, ~8 h for 1e5 configs). Falls back to plain ASE `get_potential_energy` for non-LAMMPS calculators.
   * `V2_tilde_from_z` classical branch now uses the correct per-mode coefficient `w * 0.5 * omega^2 * sigma^2` (per LatticeDynamicsToolkit's `_v2_tilde_coefficients`); classically it now equals the exact harmonic energy to machine precision. The quantum branch is algebraically unchanged.
   * `print_ethan_table` referenced `CumulantResult` fields that no longer exist; any call raised AttributeError.

3. **Integrate with post-#274 main**: restore `ForceConstants.irred_to_full` / `_build_shifted_rep` (removed on main as dead code; their caller lands here), repair the `kaldo.cumulant.__init__` exports, pin the current `cumulant_thermo` signature in the API test, and re-pin the LJ Ar F1 value at the unconverged 2^3 mesh (0.30% shift from the SNF replica-table phase-convention refinement; converged meshes unchanged).

## Testing

Full battery on this branch: **48 passed, 0 failed, 28 env-var-gated skips** (production-size fixtures), including the TDEP loader, fourth-order, and non-diagonal regression suites. CI-runnable coverage includes the Phase-5 frame regression (sampler/contractor agreement to machine precision, plus a guard that the un-permuted IFC2 visibly disagrees), the batch-evaluator contract, the V2_tilde and table fixes, and the LJ/SW cross-chemistry convergence pins.

**Remaining before merge**: the fast LAMMPS path (`_lammps_batch.py`) needs a smoke test against a real LAMMPS build (~1000 configs: energies match a plain LAMMPSlib loop, per-config time drops). No LAMMPS Python module is available in the dev environment here, so this is best run on a cluster.

Supersedes #254.

## Summary by CodeRabbit

* **New Features**
  * Added a complete cumulant thermodynamics workflow for harmonic, cubic, and quartic free-energy corrections.
  * Added Monte Carlo sampling, bootstrap uncertainty estimates, and reusable energy evaluation.
  * Added TDEP input loading, phonon thermodynamics utilities, and a stable public API.
  * Added examples for silicon, argon, and cached benchmark workflows.

* **Documentation**
  * Added usage guidance and workflow examples.

* **Bug Fixes**
  * Improved support for supercell mappings, atom-frame consistency, and non-diagonal structures.

* **Tests**
  * Added broad regression, integration, cross-chemistry, and numerical consistency coverage.


gbarbalinardo:
Superseded by #283: identical content, but with the original per-commit history replayed (authors and dates preserved) instead of a squashed module commit, on an em/ branch.

Issue #279, opened 2026-07-06
Follow-ups from QE replica-order fix: writer bug, VASP path, hex interpolation residual, NAC


gbarbalinardo:
Follow-up items surfaced while fixing the QE hexagonal replica-order bug (#272). None block that PR; grouping them here so they are not lost.

### 1. `save_second_order_qe_matrix` writes a corrupted replica index
`kaldo/interfaces/shengbte_io.py:430`:
```python
file.write('%4d %4d %4d' % (int(l_vec[2]), int(l_vec[1]), int(l_vec[2])))
```
Writes `(t3, t2, t3)`: `t1` is dropped and `t3` is duplicated. Any `espresso.ifc2` produced by kaldo has a wrong replica index except when `t1 == t3`. Fix to `(l_vec[0], l_vec[1], l_vec[2])` and verify the C/F flattening convention matches the reader (see #272). A save-then-load round-trip on a known `.fc` would make a good regression test.

### 2. VASP / ShengBTE format load path has the same C-data / F-grid pattern
`SecondOrder.load` (VASP arm) and `ThirdOrder.load` were deliberately left at `grid_type="F"` in #272 because second and third order are jointly consistent there and there is no hexagonal VASP-format validation data in the suite. The pattern is the same as the QE bug and should be validated for an anisotropic cell before changing. The AlmaBTE `GaN_wurtzite` `FORCE_CONSTANTS` + `BORN` dataset is a natural reference.

### 3. Residual hexagonal interpolation error near the A point (~35 cm^-1)
After the #272 fix, both the folded (`is_unfolding=False`) and unfolded paths still deviate from QE `matdyn.x` by up to ~35 cm^-1 for wurtzite GaN, worst near A, and split some symmetry-required degeneracies. Likely cause: replica-image assignment. The folded path wraps replicas by grid index (`Grid.grid(is_wrapping=True)`, no geometry); the unfolded path builds Wigner-Seitz weights whose candidate L-vector set / boundary tolerance may be incomplete for hexagonal supercells. `matdyn` uses proper Cartesian WS weights.

### 4. Non-analytical correction (NAC) issues for polar materials
Independent of the ordering bug (a file with NAC flag `F` never triggers NAC):
- **Silent skip:** `read_second_order_qe_matrix` sets `charges = None` when the QE NAC flag is `F` and NAC is silently disabled, with no warning. For a polar material this yields a dispersion with no LO-TO splitting that looks plausible but is wrong. Emit a warning.
[...]

PR #284, opened 2026-07-10
Declare the C replica order the vasp readers produce

gbarbalinardo:
## Motivation

The vasp-* branches of `SecondOrder.load` and `ThirdOrder.load` declared an F-ordered replica grid over data their readers produce in C order, the VASP twin of the QE mislabel fixed in #272 (item 2 of #279). The mismatch pairs replica vectors with the wrong force-constant blocks, and it is not a symmetry relabeling (labels flip without rotating Cartesian components), so it distorts results even on cubic fixtures.

`elastic_prop` carried an unconditional compensation for this (thanks @bkhli, it was doing real load-bearing work for vasp loads): whenever the declared grid was F it regenerated the replica list in C order. Correct for the vasp mislabel, but it corrupted self-consistent F loads such as hiphive, whose grid and data enumeration agree. Review iterations that kept the compensation (conditional stamps, a value-order property) all amounted to encoding an exception for vasp; the clean fix is for the loaders to declare the order their data actually uses, so no exception exists anywhere.

## Evidence that C is the data's enumeration

- **Second order**: with C pairing the vasp Si stiffness tensor satisfies the cubic equal-constant relations to machine precision (spread 7e-13) and matches the pinned `test_elastic_vasp` references; with F pairing the cubic form is destroyed (spread 1.4e2 GPa).
- **Third order**: both vasp third readers are invoked with `order='C'` and map the file's explicit lattice vectors onto a C-raveled index, the same convention #272 validated for qe-d3q.

## What changed

- vasp-* second and third order now declare `grid_type='C'` (flipped together, mirroring #272).
- `elastic_prop` trusts the loader's grid; the unconditional C-override is gone.
- The vasp-d3q Ge conductivity references are re-pinned as explicit cross-backend bands (`assert_allclose`, rtol=3e-2, centered between the observed x86-64/OpenBLAS and arm64/Accelerate values). On this cubic fixture the relabeling shift is comparable to the degenerate-subspace backend spread documented in #270, so these tests are sanity bands, not enumeration detectors; the enumeration is pinned by `test_elastic_vasp` and the new invariance test.
- New enumeration-invariance regression test in `test_elastic.py`: a C-ordered `SecondOrder` and a self-consistent F-relabeled twin of the same random value tensor on an asymmetric (2, 2, 3) supercell must give identical elastic tensors. It fails against the old unconditional override.

## Testing

Full local suite: 219 passed, 10 env-gated skips. The re-pinned bands were verified to contain both observed backend values for all three conductivity methods.

Resolves the vasp half of #279 item 2; validating the third order on an anisotropic vasp dataset (e.g. AlmaBTE GaN) remains desirable and is noted there.

### DIFF
--- a/kaldo/forceconstants.py
+++ b/kaldo/forceconstants.py
@@ -4,7 +4,7 @@ Anharmonic Lattice Dynamics
 """
 import numpy as np
 from sparse import COO
-from kaldo.grid import wrap_coordinates, Grid
+from kaldo.grid import wrap_coordinates
 from kaldo.observables.secondorder import SecondOrder
 from kaldo.observables.thirdorder import ThirdOrder
 from kaldo.observables.fourthorder import FourthOrder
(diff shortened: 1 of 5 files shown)


bkhli (approved):
`kaldo/tests/test_elastic.py` is a very strong test. I am convinced that we finally resolved the grid issue.

PR #286, opened 2026-07-11
Add pheasy force-constant importer (format='pheasy')

gbarbalinardo:
# Add a pheasy force-constant importer (format='pheasy')

## What this adds

[pheasy](https://gitlab.com/cplin/pheasy) (Lin, Han, Xu, Marzari, [arXiv:2508.01020](https://arxiv.org/abs/2508.01020)) extracts interatomic force constants from displacement-force datasets by symmetry-constrained regression, with exact invariance conditions and long-range Coulomb separation. It already exports to ShengBTE, Phono3py, Phoebe, and GPU_PBTE; this PR makes kaldo a first-class consumer:

```python
fc = ForceConstants.from_folder('pheasy_run/', supercell=(3, 3, 3), format='pheasy')
```

reads an unmodified pheasy working directory:

- primitive cell from `POSCAR` (VASP mode) or `cell.in` (QE mode)
- second order from phonopy-style `FORCE_CONSTANTS` text (compact or full header, eV/A^2) or `fc2.hdf5`
- third order from ShengBTE `FORCE_CONSTANTS_3RD` (eV/A^3) or `fc3.hdf5`
- fourth order from `FORCE_CONSTANTS_4TH` with `include_fourth=True` (the guard now accepts tdep and pheasy)
- Born effective charges and dielectric tensor from pheasy's `born.fmt`, enabling the non-analytic correction automatically

pheasy enumerates supercell atoms atom-major with the first lattice vector fastest; the importer decodes every index explicitly and maps it onto kaldo's replica grid, so no ordering assumptions are shared between the codes. kaldo never imports pheasy (BSD vs GPL); the integration is file-based only.

## Stacked on #288

Validating this importer against pheasy's real fitted outputs exposed a long-standing silent bug in kaldo's ShengBTE FC3 reader; that fix (plus the golden recalibration it requires) is split out as #288, and this PR's base is set to its branch. Only the pheasy feature commits appear in this diff; the new fourth-order parser added here resolves cell offsets through the same shared helper from the start. After #288 merges, this PR rebases onto main.

## Testing

- 40 tests in `kaldo/tests/test_pheasy_io.py`: unit tests on synthetic files in tmp_path (both row-label conventions, compact/full, text/hdf5, poisoned-redundant-row leak detectors, negative-offset conventions, truncation, validation errors) plus differential integration tests.
- Differential fixtures generated by kaldo itself (no GPL code or data enters the repo): `si-crystal/pheasy/` derives from the existing hiphive reference via a committed `make_fixture.py`; force constants agree with the hiphive route to 4e-16 under the replica permutation (negative-controlled: an identity mapping mismatches by 5 eV/A^2, since the hiphive fixture is F-ordered and pheasy files are decoded onto a C grid). `mgo/pheasy/` derives from the qe reference and checks NAC equivalence through `born.fmt` + `fc2.hdf5`.
- Cross-validated against pheasy's own shipped fitted Si output (full-format FC2, QE mode, 4x4x4): three acoustic modes at ~2e-7 THz and the triply degenerate optical mode at 15.29 THz.
- Full suite green; docs entry added to `api_forceconstants.rst` (validated against pheasy v0.1.0).

A matching worked example is proposed in nanotheorygroup/kaldo-examples (companion PR).


ZKC19940412:
Hi G. I had used pheasy myself with kALDo and its output ifc format in vasp format naturally compatible to our existing vasp-sheng parser, and it's ifcs output in qe format is naturally compatible to our qe-sheng format.  What is really missing is the mlip ifcs compatibility.

 Does the new parser addresses that? If not it is more userful to have the in draft pheasy parser also to be compatible with mlip based ifcs. 

Aother note in case cladue has not caught it. pip install pheasy arrived at very ancient pheasy version and the latest pheasy can only be imported from source via https://gitlab.com/cplin/pheasy. Definitely need to check if Claude has been using the new or the old version.

Best,

Zekun

gbarbalinardo:
Hi Zekun,
the parser reads pheasy's output files, so it does not care where the forces came from: IFCs fitted on MLIP forces load through `format='pheasy'` exactly like DFT ones. The piece that is actually missing for MLIPs is upstream, getting the forces into pheasy in the first place (its readers only parse vasprun.xml and QE output). The clean way to close that is kaldo generating the displaced supercells, evaluating them with any ASE calculator, and handing the forces to pheasy's fit directly. That is deliberately not in this PR, and it is the natural follow-up.

On vasp-sheng and qe-sheng: they do work, but you need `--full_ifc` (the compact default breaks the vasp reader), the file renames, and a hand-written CONTROL if you want the Born charges. Those routes also went through the FC3 reader that was silently dropping cell offsets outside the minimum-image range, fixed in #288, so I would rerun anything that used them before trusting the numbers. The dedicated format reads the compact default, picks up born.fmt for the NAC automatically, and adds the fourth order, which no other route reads.

You are right about pip: PyPI only has 0.0.2. I updated the docs to point at the GitLab source and pinned the commit we validated against.

gbarbalinardo:
@ZKC19940412 PR ready for review.

Issue #290, opened 2026-07-11
Gauge invariance in degenerate subspaces: one root cause behind the backend-dependent results


gbarbalinardo:
## The root cause

Numerical diagonalizers return an arbitrary orthonormal basis within each degenerate subspace, and the choice differs across BLAS/LAPACK backends (Apple Accelerate vs OpenBLAS vs MKL) and architectures. Frequencies agree to 1e-10 across machines; eigenvectors do not. Every observable that consumes mode-resolved eigenvector-dependent quantities inside degenerate clusters inherits that arbitrariness. On the small test meshes the effect is large: 25% of the modes of the Ge 3x3x3 fixture sit in exactly degenerate pairs.

This issue collects the symptoms treated one at a time so far, and tracks the combined fix.

## Symptom class A: backend-dependent transport goldens

Measured spreads between arm64/Accelerate and x86-64/OpenBLAS on identical code:

| test | spread | current state |
|---|---|---|
| test_io.py qhgk (Si) | ~0.9% (three stacks, #270) | widened to rtol=2e-2 (#281) |
| test_crystal_vasp_d3q (Ge, 3 methods) | ~2.9% | widened to rtol=3e-2 bands (#284) |
| test_crystal_qe_vasp qhgk | was failing arm64 vs x86 (#238) | pinned via fixed diffusivity_bandwidth=1.0 (#260) |
| test_ge_qe_crystal iso_bw | basis-dependent per-mode value | pinned via degenerate-triplet mean (#260) |

Latent population: ~10 more test files pin conductivities computed with the default (gauge-sensitive) adaptive broadening and no fixed bandwidths; they pass today by tolerance granularity and will flake on the next BLAS change.

Mechanism, verified on the Ge fixture: the default ShengBTE broadening builds per-pair sigma from velocity differences, which are gauge-dependent within degenerate subspaces; swapping only that ingredient moves kappa by the same magnitude as the cross-backend spread (default 0.837, tdep kernel 0.862, fixed bandwidth 0.807 on the same machine and code).

## Symptom class B: gauge-contaminated per-mode observables

- Per-mode phase_space / bandwidth / gamma-tensor entries differ across symmetry-equivalent k-points with the ShengBTE kernel, up to ~13% on Si 3x3x3 (documented in #273, which added the gauge-invariant TDEP kernel and the use_q_symmetry warning). Marginal sums are correct.
[...]

PR #293, opened 2026-07-12
Deterministic eigenvector gauge in degenerate subspaces (spike, #290)

gbarbalinardo:
Step 2 of #290, the root-cause fix. Diagonalizers return an arbitrary basis inside each degenerate cluster, and different BLAS backends pick differently, so every mode-resolved quantity inherits the choice. kaldo.gauge.canonicalize_eigenvectors removes the freedom: each cluster is rotated into the eigenbasis of a fixed operator (a graded diagonal projected into the cluster) and each phase is pinned on the largest component. The result depends only on the dynamical matrix, so two machines produce the same eigenvectors.

Opt-in via Phonons(is_canonical_gauge=True); defaults are untouched. All existing goldens pass, and the gauge-free QHGK fixed-bandwidth value is bit-identical under the flag. The unfolded path raises NotImplementedError for now, and canonical runs store under a separate cache path.

Two referees. The unit test scrambles the gauge with random unitaries inside each cluster plus random phases (the freedom another backend exercises) and asserts canonicalization recovers the reference basis to 1e-10. The cross-backend one: the Ge vasp-d3q rta and inverse values, which keep a ~2% backend spread even with fixed sigma, are pinned under the canonical gauge at rtol=5e-3 from arm64 numbers (0.647088 and 0.714601), and the x86 CI run holds them. That was the success criterion #290 set for this spike.

Making this the default would change published per-mode numbers, so that stays a separate discussion. Once merged, the two 3e-2 bands in test_crystal_vasp_d3q can be tightened.

### DIFF
--- a/kaldo/gauge.py
+++ b/kaldo/gauge.py
@@ -1,0 +1,75 @@
+"""Deterministic eigenvector gauge for degenerate subspaces.
+
+Numerical diagonalizers return an arbitrary orthonormal basis within each
+degenerate eigenvalue cluster, and the choice differs across BLAS/LAPACK
+backends. Every observable built from mode-resolved eigenvector quantities
+inside those clusters (velocities, scattering amplitudes, per-mode
+bandwidths) inherits that arbitrariness, which is the root cause tracked in
+issue #290.
+
+This module removes the freedom: within each degenerate cluster the basis is
+rotated to diagonalize a fixed generic Hermitian operator (a graded diagonal
+in the Cartesian-atom representation), whose projected spectrum is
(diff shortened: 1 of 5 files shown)


gbarbalinardo:
Closing unmerged, on purpose. The spike answered #290's question (the referee passed: the Ge cross-backend residual is pure gauge and collapses below 5e-3 under a canonical basis), but shipping the flag adds a public knob that changes per-mode numbers without making them more physical, to tighten two bands on one fixture. Not worth owning. The recipe and validation live in this branch and are summarized on #290.

PR #296, opened 2026-07-12
Fix third-order scratch assembly: OOM on full-range tensors and data loss on failure

gbarbalinardo:
`_assemble_from_scratch_third` deleted each chunk file as it read it, before the tensor existed, so any failure past that point destroyed the computed forces. It also built the tensor by reshaping a 5-D COO, whose flat index (n_atoms*3)*N3*N3 overflows int32 and spawns several nnz-sized int64 temporaries; on a full-range (unthresholded) 216-atom cell that peaked past 60 GB and was OOM-killed, losing hours of MatterSim forces.

The fix assembles the (n_atoms*3, N3, N3) COO directly from the chunk coordinates: the two output rows are iat*3+ic and jat*3+jc, computed in int64, and the k axis is copied as is. No 5-D reshape, no dense intermediate. Scratch files are removed only after the tensor is built.

Verified bitwise-identical to the previous tensor on a 216-atom a-Si cell (272M nonzeros, max abs difference 0), peak memory 28 GB against the previous kill, 108 s.

Tests: assembly matches a dense reference, scratch survives a failed assembly, scratch is removed on success.

## Summary by CodeRabbit

* **Performance**
  * Improved third-order tensor assembly from scratch data to be more memory-efficient and faster.
* **Reliability**
  * Refined scratch cleanup behavior: scratch files are preserved until the final assembly completes, and are removed only after successful completion when cleanup is enabled.
  * If assembly fails, scratch data remains available for recovery.
* **Documentation**
  * Clarified `keep_scratch` semantics for third-order assembly.
* **Tests**
  * Added regression coverage for correct assembly against a dense reference and for cleanup behavior on both success and failure.

### DIFF
--- a/kaldo/controllers/displacement.py
+++ b/kaldo/controllers/displacement.py
@@ -664,8 +664,9 @@ def calculate_third(atoms, replicated_atoms, third_order_delta, distance_thresho
         are accumulated in memory (original behaviour).
     keep_scratch : bool
         If True, scratch files are preserved after successful assembly. If False
-        (default), each file is deleted as it is consumed during assembly, and the
-        directory is removed if empty.
+        (default), scratch files and sentinels are removed only after the tensor
+        is fully built (a failure mid-assembly never destroys computed forces),
+        and the directory is removed if empty.
     jat_flush_every : int
         Number of jat iterations to buffer before flushing to disk. Only used
         when ``scratch_dir`` is set. Default 50.
--- a/kaldo/controllers/displacement.py
+++ b/kaldo/controllers/displacement.py
@@ -956,38 +956,56 @@ def _flush_chunk_third(scratch_dir, iat, chunk_id, chunk_coords, chunk_values):
 
 
 def _assemble_from_scratch_third(scratch_dir, n_atoms, n_replicas, keep_scratch):
-    """Build the final COO tensor from per-jat-chunk scratch files using two passes.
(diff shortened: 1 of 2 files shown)


gbarbalinardo:
Note while validating on the original 216-atom case: the data-loss half of this (scratch preserved until the tensor is built) is confirmed working. The memory half needs more, the dense route's COO.from_numpy still spikes on a fully-dense unthresholded tensor because np.nonzero allocates several nnz-sized index temporaries and the 5-D to 3-D COO.reshape linearizes an index that overflows int32. I am testing a direct scipy CSR construction (compute the two flat output axes in int64, hand to coo_matrix, no dense intermediate) and will update this branch with the version that actually holds at 216 atoms. Holding the memory claim until then; the ordering fix stands on its own.

gbarbalinardo:
Updated: the memory caveat from my earlier comment is resolved. The dense-array route I first pushed still spiked on a fully-dense tensor (COO.from_numpy runs np.nonzero on a near-full array), so I replaced it with a direct build of the output-space COO from the chunk coordinates, no dense intermediate and no 5-D reshape. Verified bitwise-identical to the prior tensor on the real 216-atom a-Si case, peak 28 GB where the old path was OOM-killed. Branch squashed to a single commit.

niklundgren (approved):
Looks like a solid update to testing and third order assembly algorithm (especially avoiding the large initializations used in previous version).

PR #299, opened 2026-07-19
NAC bugs in `HarmonicWithQ`

bkhli:
Two bugs of NAC here:

1. FCC-only half-phase in `nac_dynmat` (frequencies).** The
   reciprocal-space Ewald dipole phase is evaluated at *half* its correct value
   for every non-FCC Bravais lattice, so LO/TO frequencies of non-cubic polar
   crystals are wrong.
2. Ewald parameter `Lambda` mismatch between `nac_dynmat` (frequencies)
   and `nac_derivatives` (group velocities).** The velocity operator uses a
   different Ewald `Lambda` than the dynamical matrix, so it is not the q-gradient
   of the matrix that goes into the frequencies; NAC group velocities are wrong
   for cells whose first lattice vector is not axis-aligned (e.g. FCC primitive).

@niklundgren Please check it out if your new code branch fixed this issue or not. I have provided the test case and you can check it out. 

---

### Root cause
`nac_dynmat` builds the Ewald grid in a *normalized* unit system and applies a
phase prefactor of `pi`:

- `harmonic_with_q.py:527` — `positions_n = atoms.positions / atoms.cell[0,:].max()`
- `harmonic_with_q.py:529-530` — `reciprocal_n = inv(cell); reciprocal_n /= |reciprocal_n[0,0]|`
- `harmonic_with_q.py:580` / `:595` — `phase = exp(1j * pi * g_positions·distances_n)` (and the `cos` form at Γ)

The applied phase is `exp(i·[pi/(c_recip·a_max)]·G·Δr)`; the physical phase is
`exp(i·2π·G·Δr)`. These are equal **iff `c_recip·a_max = 0.5`**, an accident of the
FCC primitive cell. For simple-cubic / tetragonal / orthorhombic / hexagonal
`c_recip·a_max = 1.0`, so exactly **half** the phase is applied. The correct
convention already existed in the sibling `nac_derivatives`.

### The fix
Only the phase prefactor changes (`pi` → `2*pi*a_max*c_recip`), which equals the
physical `G·r` built from `2π·inv(cell/Bohr)` reciprocal vectors and Bohr
distances. **Bit-for-bit identical for FCC primitive cells** (`c_recip·a_max=0.5`
→ prefactor `pi`); restores the correct phase for all other lattices.

### Measured
- **FCC/MgO unchanged**: dynamical-matrix element delta ≤ 1.1e-13; MgO still
  matches QE `matdyn.x` to **0.00 cm⁻¹**.
- **Symmetry restored**: the C6v-required Γ–A transverse-acoustic degeneracy of
  wurtzite (GaN) goes from a **36.5 cm⁻¹** split (wrong) to **0.0 cm⁻¹**.

### Root cause
`nac_dynmat` and `nac_derivatives` both evaluate the reciprocal-only Ewald dipole
sum, whose value depends on the Ewald splitting parameter `Lambda` (it sets the
Gaussian decay `exp(-geg/(4*Lambda))/geg`, the reciprocal cutoff
`geg/(4*Lambda) < gmax`, and the reciprocal-grid extent `geg0 = 4*Lambda*gmax`).
The two routines used **different** `Lambda`:

| routine | used for | `Lambda` | line |
|---|---|---|---|
| `nac_dynmat` | frequencies | normalises reciprocal grid by `c_recip=\|inv(cell)[0,0]\|`, `Lambda = 1` | `:524` |
| `nac_derivatives` | group velocities | `(2*pi*Bohr/\|cell[0,:]\|)**2` | `:653` (`:643` pre-`0001`) |

These give the same physical Gaussian width only when
`|cell[0,:]| * |inv(cell)[0,0]| == 1`, i.e. when the first lattice vector is
axis-aligned. That holds for the wurtzite fixture (so its velocity `Lambda` was
accidentally correct) but **not** for an FCC primitive cell (MgO:
`|cell[0,:]|·|inv(cell)[0,0]| = 0.707 ≠ 1`). There `nac_derivatives` is **not**
the q-gradient of `nac_dynmat`.

`nac_dynmat`'s `Lambda = 1` is the validated one — it reproduces QE `matdyn.x`
for MgO to 0.00 cm⁻¹ (VALIDATION §1). Folding `nac_dynmat` onto the velocity
`Lambda` instead would move MgO frequencies **26–43 cm⁻¹** off `matdyn.x`, so the
frequency path must be preserved and the **velocity** path corrected.

### The fix
Introduce a shared helper `_ewald_reciprocal_scale(cell)` returning
`|inv(cell)[0,0]|`, used by **both** routines, and set `nac_derivatives`'s Ewald
parameter to `Lambda = (2*pi*c_recip*Bohr)**2` — the same physical Gaussian width
`nac_dynmat` already uses. `nac_derivatives` then becomes the exact analytic
gradient of `nac_dynmat` (finite-difference residual ~8% → <0.01%). The shared
helper means the two routines cannot drift apart again.

`nac_dynmat`'s numerical output is **bit-identical** (it sources `c_recip` from the
helper, whose value equals the previous inline expression), so **no phonon
frequency changes**; only NAC **group velocities** change.

### Measured
- **Gradient consistency** `nac_derivatives == 1j·d/dq[nac_dynmat]`, reference-free
  finite difference on the FCC/MgO fixture: **8.36% → 0.007%** residual.
- **Frequencies unchanged by `0002`**: MgO Δ = 8.5e-14 cm⁻¹, GaN Δ = 0.000 cm⁻¹
  (second bug fix never touches the frequency path).
- **Velocity operator change (the effect)**: MgO NAC velocity operator moves
  **3.67%**; larger for cells further from `|cell[0,:]|·|inv(cell)[0,0]| = 1`.

## Behavior change for users

Across the two commits, for **non-FCC polar** materials:

- **Frequencies shift** — caused by **`0001`** (the phase fix). Non-cubic polar
  LO/TO frequencies move (tens of cm⁻¹). FCC frequencies are unchanged.
- **Group velocities (and thus κ) shift** — caused by **`0001` (phase)** for
  non-FCC polar cells **and** by **`0002` (Ewald `Lambda`)** for any polar cell
  whose first lattice vector is not axis-aligned (including **FCC**, e.g. MgO
  ~few %).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved non-analytic correction calculations for crystals with varied cell orientations and non-FCC geometries.
  * Corrected reciprocal-space phase handling for finite-q and Gamma-point calculations.
  * Aligned NAC dynamical-matrix and derivative calculations for more consistent phonon velocities.

* **Tests**
  * Added regression coverage for hexagonal-cell degeneracies, active NAC frequency shifts, and existing FCC reference behavior.

### DIFF
--- a/kaldo/observables/harmonic_with_q.py
+++ b/kaldo/observables/harmonic_with_q.py
@@ -16,6 +16,16 @@ logging = get_logger()
 
 MIN_N_MODES_TO_STORE = 1000
 
+
+def _ewald_reciprocal_scale(cell):
+    """Reciprocal length |inv(cell)[0, 0]| that sets the Ewald Gaussian width of
+    the non-analytic correction. Shared by ``nac_dynmat`` (which normalises its
+    reciprocal grid by it, with Lambda = 1) and ``nac_derivatives`` (which sets
+    its Ewald parameter Lambda from it) so that the NAC group-velocity operator
+    stays the exact q-gradient of the NAC dynamical matrix. See the note in
+    ``nac_derivatives``."""
+    return np.abs(np.round(np.linalg.inv(cell), 12)[0, 0])
+
 _warned_incommensurate = False
 
 
@@ -528,7 +538,7 @@ class HarmonicWithQ(Observable, Storable):
         positions_n = atoms.positions.copy() / a_max  # Normalized positions
         distances_n = positions_n[:, None, :] - positions_n[None, :, :]  # distance in crystal coordinates
         reciprocal_n = np.round(np.linalg.inv(atoms.cell), 12)  # round to avoid accumulation of error
-        c_recip = np.abs(reciprocal_n[0, 0])  # length used to normalize reciprocal vectors
+        c_recip = _ewald_reciprocal_scale(atoms.cell)  # length used to normalize reciprocal vectors (shared with nac_derivatives)
         reciprocal_n /= c_recip  # Normalized reciprocal cell
         # Prefactor for the physical G.r phase (see phase factor below). g_positions and
         # distances_n are built in the normalized unit system only to set the Ewald grid and
@@ -650,7 +660,15 @@ class HarmonicWithQ(Observable, Storable):
         if gmax==None:
             gmax = 14  # maximum reciprocal vector (same default value in ShengBTE/QE)
         if Lambda==None:
-            Lambda = (2*np.pi*units.Bohr/np.linalg.norm(cell[0,:]))**2  # Ewald parameter
+            # Ewald parameter. Must reproduce the SAME physical Gaussian width as
+            # nac_dynmat, which normalises its reciprocal grid by
+            # c_recip = |inv(cell)[0, 0]| and uses Lambda = 1. Otherwise
+            # nac_derivatives is not the q-gradient of nac_dynmat and the NAC
+            # group velocities are wrong for cells whose first lattice vector is
+            # not axis-aligned (e.g. FCC primitive: ~8%). The historical form
+            # (2*np.pi*units.Bohr/np.linalg.norm(cell[0,:]))**2 coincides with
+            # this only when |cell[0,:]| * |inv(cell)[0, 0]| == 1.
+            Lambda = (2*np.pi*_ewald_reciprocal_scale(cell)*units.Bohr)**2  # Ewald parameter
         geg0 = 4 * Lambda * gmax
         omega_bohr = np.linalg.det(atoms.cell.array / units.Bohr) # Vol. in Bohr^3
         positions_bohr = atoms.positions.copy() / units.Bohr
--- a/kaldo/observables/harmonic_with_q.py
+++ b/kaldo/observables/harmonic_with_q.py
@@ -524,10 +524,20 @@ class HarmonicWithQ(Observable, Storable):
             Lambda = 1 # (2*np.pi*units.Bohr/np.linalg.norm(atoms.cell[0,:]))**2
         geg0 = 4 * Lambda * gmax
         omega_bohr = np.linalg.det(atoms.cell.array / units.Bohr) # Vol. in Bohr^3
-        positions_n = atoms.positions.copy() / atoms.cell[0, :].max()  # Normalized positions
+        a_max = atoms.cell[0, :].max()  # length used to normalize real-space distances
+        positions_n = atoms.positions.copy() / a_max  # Normalized positions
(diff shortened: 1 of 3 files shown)


bkhli:
I have tested these patches results on phonondb results and found out that it improves the numerical consistency with phono3py results. Current NAC code has a large error compared to phono3py.

gbarbalinardo (changes requested):
@bkhli Nice catch on the `pi` prefactor. Both bugs are real, and I like that the fix is bit-for-bit identical on FCC so MgO frequencies don't move. Checked the phase identity on MgO, GaN and a few random cells, holds to ~1e-11.

A few things before merging.

1. `test_nac_qe.py::test_nac_velocity` fails on this branch, so CI will be red. The committed reference is `[33.906, 33.906, 32.844, 8.682, 8.682, 21.596]` and you produce `[29.512, 29.512, 46.846, 9.993, 9.993, 32.476]`. Central finite differences of the NAC frequencies give `[29.505, 29.505, 46.817, 9.989, 9.989, 32.492]`, so your values are right and the old reference came from the buggy code. Please update it here, with a comment saying why. Please don't widen the tolerance.

2. The description says the velocity operator moves 3.67%, but the group velocity norms move up to 43% (`32.84` -> `46.85`). That's what users see and it changes published kappa for polar materials. Can you update the description?

3. `n_greplicas` still uses `axis=0` at line 568 and `axis=1` at line 693. Since `contract('ib,ab->ia', g_replicas, reciprocal)` puts the vector for `b` in column `b`, `axis=0` looks correct. GaN gives `[14,14,26]` vs `[16,13,26]`; MgO hides it because it's cubic. Same class of bug this PR fixes, can we do it here?

4. The gradient test is MgO only, so it can't catch (3). Please add it on the GaN fixture too.

5. Minor: `_ewald_reciprocal_scale` returns `|inv(cell)[0,0]|`, which collapses to zero for a cell whose first reciprocal row has no x component. Inherited, not yours, but now that it's a named helper please add a guard.

6. Minor style: `import tensorflow as tf` at the top of the test file, `np.testing.assert_allclose` with explicit `rtol`/`atol` instead of `abs(...) < tol`, `np.float64` instead of `float`. Please also drop the "(Found and fixed by Claude Opus 4.8)" lines from the commits.

The reference-free gradient test is a good idea, it can't rot against a stale reference, which is exactly what happened in (1). On that: can you check whether the other NAC references have the same provenance?

Note this overlaps with #262, which deletes `nac_dynmat` and `nac_derivatives`. I'd like to merge yours first since it fixes what's in main today.

bkhli:
The bug is corrected in #303. #303 is the intended way.

PR #301, opened 2026-07-25
Fix phonon frequencies at incommensurate q on non-diagonal tdep supercells

### LINKED ISSUE #297: Wrong phonon frequencies for non-diagonal (SNF) supercells with an odd/3-fold Smith-normal-form factor
phonopy is correct here but kaldo gives imaginary phonon. 

---

## Summary

When a second-order force-constant set is ingested through kALDo's
`format='tdep'` reader for a **non-diagonal** primitive→supercell transform whose
Smith normal form contains an **odd / 3-fold factor**, `HarmonicWithQ` returns
phonon frequencies that are wrong by up to **~85 cm⁻¹**, with the error
concentrated in the **acoustic branches** and vanishing exactly at Γ. The same
force constants, in the same files, are reproduced to **< 1 cm⁻¹** by phonopy /
phono3py and match the reference frequencies computed by ALAMODE. A *different*
non-diagonal supercell of the same primitive family — an FCC-type det-32 cell —
is reproduced by kALDo **exactly (0.000 cm⁻¹)**, so the failure is specific to the
supercell shape (its SNF), not to non-diagonality per se and not to the input FCs.

## Environment

| package | version |
|---|---|
| kALDo | **2.2.1+15.g3544a5f** (latest git) |
| python | 3.12.13 |
| numpy | 2.4.6 |
| ase | 3.22.1 |
| tensorflow | 2.19.1 |

## Steps to reproduce

```bash
python repro.py        # in the kaldo env above
```

For each material it:

1. loads `data/<mp_id>/infile.{ucposcar,ssposcar,forceconstant}` via
   `ForceConstants.from_folder(..., format='tdep', supercell_matrix=P, only_second=True)`;
2. evaluates `HarmonicWithQ(q_point=q, second=fc.second).frequency` at the
   reference q-points in `data/<mp_id>/expected.json`;
3. prints observed vs expected frequencies (cm⁻¹) and the max deviation;
4. exits **nonzero** if any material deviates by > 1.5 cm⁻¹ away from Γ.

The two bundled cases are:

| mp_id | supercell matrix P (rows, in primitive units) | det | non-diagonal | result |
|---|---|---|---|---|
| mp-1000 | `[[-2,2,2],[2,-2,2],[2,2,-2]]` (FCC) | 32 | yes | **PASS** (0.000 cm⁻¹) |
| mp-7 | `[[1,-1,0],[0,1,-1],[3,3,3]]` | 9 | yes | **FAIL** (85.1 cm⁻¹) |

## Observed vs expected

`repro.py` output (abridged):

```
=== mp-1000: non-diagonal supercell, det=32, n_prim=2, n_sc=64 ===
  ... every q: max|obs-exp| = 0.000 cm^-1
  --> max deviation away from Gamma: 0.000 cm^-1 (PASS)

=== mp-7: non-diagonal supercell, det=9, n_prim=6, n_sc=54 ===
  q=[0. 0. 0.]              max|obs-exp|=  0.036 cm^-1
  q=[0.6667 0.3333 0.1111]  max|obs-exp|= 81.729 cm^-1  <-- MISMATCH
      observed : [-43.1  49.9  70.9 ...]
      expected : [ 38.6  56.3  62.1 ...]
  q=[0.3333 0.6667 0.2222]  max|obs-exp|= 85.107 cm^-1  <-- MISMATCH
      observed : [-43.3  46.9  54.3 ...]
      expected : [ 41.8  53.2  71.8 ...]
  ...
  --> max deviation away from Gamma: 85.107 cm^-1 (FAIL)
```

Key features of the mp-7 error:

- **Exact at Γ** (0.036 cm⁻¹) — the on-site/ASR part is fine.
- **Acoustic branches wrong** (the three lowest modes swing by tens of cm⁻¹ and
  even go *negative/imaginary*, e.g. −43 cm⁻¹ where the reference is +39); the
  optical branches are close (a few cm⁻¹).
- **Discriminating contrast:** the FCC det-32 non-diagonal cell (mp-1000) is
  exact, so the bug is triggered by the supercell's SNF shape (det-9 here has a
  3-fold invariant factor from the `[3,3,3]` row), not by non-diagonality alone.

## What we ruled out (the input force constants are correct)

- **Same FCs → phonopy/phono3py are correct.** The identical fc2 (converted from
  ALAMODE, in eV/Å²) reproduces the ALAMODE `S.evec` reference to **0.79 cm⁻¹**
  for mp-7 and **0.000 cm⁻¹** for mp-1000 through phonopy's dynamical matrix.
  Only kALDo's tdep path is off.
- **kALDo loads the FCs correctly.** kALDo's assembled second-order tensor
  (`fc.second.value`, shape `(1, n_prim, 3, n_rep, n_prim, 3)`) equals an
  independently-built reference tensor (the same fc2 placed into kALDo's own SNF
  replica slots) to **1e-15**. So the array content and the replica assignment
  are right; the discrepancy appears only when `HarmonicWithQ` Fourier-transforms
  it.
- **Not a writer artifact.** Emitting one entry per neighbour with the full value
  vs. splitting a boundary neighbour into its M equidistant smallest-vector
  images (value/M) gives the *same* wrong result (both 85.1 cm⁻¹), so it is
  downstream of the `infile.forceconstant` content.
- **`is_unfolding=True` is worse** (~534 cm⁻¹ at Γ), so the plain path is the
  relevant one.
- **Validation subtlety (why this can hide):** the harmonic check must be done at
  specific q-points against a reference. Sampling a Monkhorst–Pack grid with
  `Phonons(kpts=(n,n,n))` and only checking `min(frequency) > 0` masks the issue
  for some cells and *manufactures* spurious imaginary modes for others (it
  samples q off the reference set), so it is not a reliable detector. `repro.py`
  compares frequencies q-by-q against the ALAMODE reference.

## Suspected locus (hypothesis)

Best guess from reading the source; not verified by a fix.

- `kaldo/observables/harmonic_with_q.py`, **`calculate_dynmat_fourier`**
  (def at line 305; the non-Γ branch at lines ~340–343) builds the dynamical
  matrix as
  `contract('ialjb,l->iajb', dynmat[0], chi(q_point, list_of_replicas, cell_inv))`
  — i.e. it applies **one phase per replica index `l`**,
  `chi = exp(2πi q·R_l)` with `R_l = list_of_replicas[l]`, **uniformly** to all
  atom pairs `(i, j)` stored in that replica. There is no per-atom-pair
  minimum-image / smallest-vector treatment (the multiplicity split that
  phonopy/phono3py apply when building `D(q)`).
- `list_of_replicas` for the non-diagonal path comes from the SNF mapping in
  `kaldo/interfaces/tdep_io.py`, **`build_supercell_replica_mapping`** (def line
  33), which **re-wraps** the replica table to Cartesian-norm-minimal vectors
  (loop starting line 109). The reader `parse_tdep_forceconstant` (line 255)
  assigns each force-constant entry to a single replica via
  `rep_id = int(rep_ids[0])` (line 346).

For a supercell whose SNF is anisotropic / has an odd factor (det-9, 3-fold),
the single canonical `R_l` used for the phase is not the correct per-pair lattice
vector for every atom pair folded into replica `l`, so `exp(2πi q·R_l)` carries
the wrong phase for part of the block — dominating the acoustic (long-wavelength)
response. Isotropic even-factor cells (FCC det-32) happen to have `R_l` coincide
with the correct per-pair vector, so they are exact. A fix would likely require
either the per-atom-pair minimum-image/multiplicity handling in
`calculate_dynmat_fourier`, or a replica-table/phase convention in
`build_supercell_replica_mapping` that stays consistent with the phase used
downstream.

## Attribution

The bundled force constants and reference frequencies (`data/mp-7/`,
`data/mp-1000/`) are derived from https://phonix-db.org,
Materials-Project entries **mp-7** and **mp-1000**. 
The `infile.forceconstant` files are unit-converted (Ry/Bohr² →
eV/Å²) from the shipped ALAMODE `FC2.xml`; `expected.json` frequencies are the
shipped ALAMODE `S.evec` eigenvalues converted to cm⁻¹. If this report is filed
or redistributed, retain this attribution.
```
data/
  mp-7/     infile.ucposcar  infile.ssposcar  infile.forceconstant  expected.json
  mp-1000/  infile.ucposcar  infile.ssposcar  infile.forceconstant  expected.json
```

[kaldo-bug-non-diagonal.zip](https://github.com/user-attachments/files/30076000/kaldo-bug-non-diagonal.zip)

gbarbalinardo:
Fixes #297.

The tdep SNF loader collapsed every IFC2 entry of a congruence class onto a single replica vector for the Fourier phases. That is exact at ssposcar-commensurate q, which is why the det-32 FCC case in the issue passes, but it interpolates badly in between: on the det-9 cell the acoustic branches came out wrong by up to 85 cm^-1, some imaginary, while phonopy reproduces the same files to 0.79 cm^-1. TDEP writes a minimum-image lattice vector for every pair, and phonopy/ALAMODE phase each pair with its own vector; we were throwing that information away.

This stores the second order on the unique per-pair vectors read from the file instead, so the plain path computes the per-pair Fourier sum. On the issue's repro bundle mp-7 goes from 85.1 to 0.76 cm^-1 against the ALAMODE reference and mp-1000 stays exact (0.000).

A few things that surfaced along the way: [0, 0, 0] is forced to replica 0, since acoustic_sum_rule targets slot 0 as the home cell and np.unique would put a negative vector there (a misplaced correction cancels at Gamma and is wrong at every other q). is_unfolding now raises on the SNF path instead of returning garbage; it assumes a diagonal tiling, and the per-pair table already is the minimum-image interpolant it would reconstruct. The det(M) class table stays attached as metadata, and third/fourth order are unchanged. NonDiagonalGrid gained an exact-row dict lookup so parsing big IFC2 files stays linear.

New tests: a self-contained det-9 model pinning the plain path against a per-pair reference at incommensurate q, an ASR placement test with a deliberate sum-rule defect (both fail on main), and the unfolding guard. E1/E2 are updated to the new contract. If you have storage='numpy' runs on non-diagonal cells, regenerate them: cached per-q artifacts predate the fix.

## Summary by CodeRabbit

* **New Features**
  * Improved non-diagonal lattice-vector resolution using exact-match-first replica lookups.
  * Added TDEP IFC2 scanning to read, round, deduplicate, and deterministically order per-pair replica lattice vectors.
  * Non-diagonal TDEP workflows now consistently use the selected per-pair replica table for interpolation setup.

* **Bug Fixes**
  * Prevents unsupported unfolding on the SNF/non-diagonal path by raising a clear error.

* **Tests**
  * Updated non-diagonal expectations and added coverage for incommensurate q behavior, unfolding rejection, and acoustic sum-rule targeting (Issue `#297`).

### DIFF
--- a/kaldo/grid.py
+++ b/kaldo/grid.py
@@ -109,22 +109,26 @@ class Grid:
 class NonDiagonalGrid(Grid):
     """Grid for a non-diagonal primitive-to-supercell tiling.
 
-    Stores an explicit ``replica_table`` (n_rep × 3) integer lattice vectors
-    in the primitive basis, already minimum-image-wrapped inside the
-    Wigner-Seitz cell of the supercell. Mimics the :class:`Grid` public
-    interface so downstream code (ForceConstant, HarmonicWithQ) can use it
-    interchangeably.
+    Stores an explicit ``replica_table`` of integer lattice vectors in the
+    primitive basis. The table holds either one norm-minimal representative
+    per Born-von-Karman congruence class (``det(M)`` rows) or, for
+    second-order TDEP loads since issue #297, one row per distinct per-pair
+    lattice vector found in the force-constant file (several rows of the
+    same class). Mimics the :class:`Grid` public interface so downstream
+    code (ForceConstant, HarmonicWithQ) can use it interchangeably.
 
     The ``is_wrapping`` flag on ``grid()`` and ``grid_index_to_id()`` is
     accepted for API compatibility with :class:`Grid` but is **always
-    treated as True**: the replica table is already minimum-image-wrapped
-    by construction.
+    treated as True**: the stored table is returned as-is.
     """
 
     def __init__(self, replica_table, M):
         self._replica_table = np.asarray(replica_table, dtype=int)
         self._M = np.asarray(M, dtype=float)
         self.grid_size = len(self._replica_table)
+        # Exact-row lookup: the dominant path when parsing per-pair IFC
+        # files (issue #297); misses fall back to congruence wrapping.
+        self._row_index = {tuple(int(x) for x in r): i for i, r in enumerate(self._replica_table)}
         # Synthesize a shape for the Grid base-class interface (used only
         # for n_replicas bookkeeping downstream).
         self.grid_shape = (self.grid_size, 1, 1)
@@ -134,7 +138,7 @@ class NonDiagonalGrid(Grid):
         return self._replica_table.copy()
 
     def grid(self, is_wrapping: bool = True):
-        # replica_table is already the minimum-image wrap
+        # the stored table is returned as-is
         return self._replica_table.copy()
 
     def unitary_grid(self, is_wrapping: bool = True):
@@ -143,15 +147,17 @@ class NonDiagonalGrid(Grid):
 
     def grid_index_to_id(self, cell_idx, is_wrapping: bool = True):
         cell_idx = np.asarray(cell_idx).astype(int)
-        if is_wrapping:
-            idx = wrap_lattice_vector_to_replica(
-                cell_idx, self._replica_table, self._M,
-            )
-            if idx < 0:
-                return np.array([], dtype=int)
-            return np.array([idx], dtype=int)
-        mask = (self._replica_table == cell_idx).all(axis=1)
-        return np.argwhere(mask).flatten()
+        hit = self._row_index.get(tuple(int(x) for x in cell_idx))
+        if hit is not None:
+            return np.array([hit], dtype=int)
+        if not is_wrapping:
+            return np.array([], dtype=int)
+        idx = wrap_lattice_vector_to_replica(
+            cell_idx, self._replica_table, self._M,
+        )
+        if idx < 0:
+            return np.array([], dtype=int)
+        return np.array([idx], dtype=int)
 
     def id_to_grid_index(self, id: int):
         return self._replica_table[int(id)].copy()
@@ -176,12 +182,15 @@ class NonDiagonalGrid(Grid):
 def wrap_lattice_vector_to_replica(R_prim_int, replica_table, M, tol=1e-4):
     """Find the replica index of a lattice vector R (integer primitive basis).
 
-    Wraps R through the non-diagonal supercell PBC and looks up the unique
-    replica entry. ``replica_table`` may be in either ``[0, 1)``-fractional
-    form or "Cartesian-norm-minimal" form; we test both candidates via
-    (a) the sc-fractional ``[0, 1)`` wrap and (b) nearby integer shifts of
-    M that could produce a norm-minimal representative. Returns the index
-    into ``replica_table`` or ``-1`` if no match.
+    ``replica_table`` may hold one row per congruence class (in
+    ``[0, 1)``-fractional or Cartesian-norm-minimal form) or several
+    congruent rows (per-pair tables, issue #297). An exact match on ``R``
+    wins over everything else, so per-pair tables resolve each vector to
+    its own row; only then the ``[0, 1)`` wrap and nearby integer shifts
+    of M are tried. For deduplicated class tables the priority is
+    irrelevant (at most one row is congruent to R), so their behavior is
+    unchanged. Returns the index into ``replica_table`` or ``-1`` if no
+    match.
 
     Pure SNF math — no TDEP-format dependencies. Used internally by
     :class:`NonDiagonalGrid` and by the TDEP non-diagonal IFC parsers.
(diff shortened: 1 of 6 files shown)


ejmeitz:
Harmonic thermodynamic properties match for SW silicon at 100K in my check. Re-running entire cumulant workflow as well.

ejmeitz:
This breaks `irred_to_full` for me. I do not have a great understanding of why.

gbarbalinardo:
Closing as included: #283 carried this fix byte-identically (the per-pair Fourier-phase changes to grid.py, tdep_io.py and harmonic_with_q.py forward-ported as ed8c582d, plus Ethan's irred_to_full class-table projection on top) and merged today, so the content is on main. The branch stays for reference.

PR #303, opened 2026-08-08
Full fix on NAC gonze for both QE and phonopy

bkhli:
This PR built on top of #262 nwl/gonze_nac_refactor. This tries to match the exact behavior of QE 7.6 and phonopy 2.17.1. 
It provides a two sets of unit tests: 

- 26 solids covering 13 Bravais lattice from phonondb calculated with phonopy 2.17.1
- 14 different bravais lattice solids with reproducible scripts that reference value of q2r.x from QE 7.6

Now, `is_nac=True` is not needed. kaldo automatically detected the correct format, read charge, and apply NAC for QE and VASP/phonopy correctly. `is_unfolding` flag historically had some connections with `is_nac`, but to load NAC correctly, no longer needs  `is_unfolding` flag.

The user-facing NAC options are limited to:

- `HarmonicWithQ(..., nac_q_direction=(1, 0, 0))` for the physical direction
  of the exact-Gamma limit. Finite-q calculations use q itself. A `Phonons`
  grid currently uses the default Gamma direction; use `HarmonicWithQ` to
  compare different directional Gamma limits.
- `nac_bvk_supercell_matrix=...` to state the defining force-constant
  Born--von Karman cell when it cannot be inferred as
  `diag(second.supercell)`. It does not request remeshing, and should normally
  be omitted.

---

## Full PR report: long-range electrostatics in harmonic phonons
The general idea is

```math
D_{\mathrm{full}}(\mathbf q;\hat{\mathbf q}) = \mathrm{FT}[\mathrm{IFC}_{\mathrm{short}}](\mathbf q) + D_{\mathrm{dipole}}(\mathbf q;\hat{\mathbf q}).
```

The first term is the ordinary Fourier interpolation of short-range interatomic force constants. The second is the long-range dipole--dipole dynamical matrix constructed from the Born effective charges and dielectric tensor; its Gamma-point limit can depend on the approach direction $`\hat{\mathbf q}`$. VASP/Phonopy and QE differ only in how $`\mathrm{IFC}_{\mathrm{short}}`$ is obtained and which matching $`D_{\mathrm{dipole}}`$ is restored.

## Why this PR is needed

In a nonpolar crystal, harmonic phonons can usually be interpolated by Fourier transforming interatomic force constants from a finite supercell. In a polar crystal, that construction is incomplete. Displacing an ion produces a macroscopic electric field through its Born effective-charge tensor, and the field is screened by the high-frequency dielectric tensor. The resulting dipole--dipole interaction is long-ranged, so it cannot be represented reliably by a finite real-space force-constant cutoff alone.

The long-range contribution has a direction-dependent limit as $`\mathbf q`$ approaches the zone centre. A longitudinal optical displacement creates a macroscopic electric field, whereas a transverse optical displacement does not create the same field. This produces longitudinal-optical/transverse-optical (LO--TO) splitting near Gamma. The direction supplied at Gamma is therefore physical input, not a numerical workaround.

This long-range reconstruction is commonly called the *non-analytic correction* (NAC). In this PR, NAC means the complete, convention-consistent separation and restoration of the dipole contribution, not merely a small frequency adjustment at Gamma.

## Direct comparison

Let $`\Phi(\mathbf R)`$ denote real-space interatomic force constants, where $`\mathbf R`$ is a lattice translation. Let $`\mathbf q_m`$ denote the q-point mesh commensurate with the force-constant supercell, and let $`\hat{\mathbf q}`$ denote the unit direction of the $`\mathbf q\to\Gamma`$ limit, used only when evaluating Gamma. The symbol $`\delta_{\mathbf q,\Gamma}`$ is an indicator equal to one at Gamma and zero elsewhere; it is not a Dirac delta. Atom and Cartesian indices, together with the usual $`1/\sqrt{M_iM_j}`$ mass weighting, are omitted below so that the two data conventions can be compared directly.

```math
\begin{aligned} D_{\mathrm{VASP}}(\mathbf q;\hat{\mathbf q}) &= \mathrm{FT}_{\mathrm{WS}}\!\left[ \Phi_{\mathrm{VASP}}^{\mathrm{total,SC}}(\mathbf R) - \mathrm{IFFT}_{\mathbf q_m}\! [D_{\mathrm{Gonze}}^{\mathrm{dipole}}(\mathbf q_m)] \right](\mathbf q) + D_{\mathrm{Gonze}}^{\mathrm{dipole}} (\mathbf q;\hat{\mathbf q}), \\[8pt] D_{\mathrm{QE}}(\mathbf q;\hat{\mathbf q}) &= \mathrm{FT}_{\mathrm{q2r}}\!\left[ \Phi_{\mathrm{q2r}}^{\mathrm{short}}(\mathbf R) \right](\mathbf q) + D_{\mathrm{QE}}^{\mathrm{rigid}}(\mathbf q) + \delta_{\mathbf q,\Gamma} D_{\mathrm{QE}}^{\mathrm{directional}}(\hat{\mathbf q}). \end{aligned}
```

| Formula term | VASP/Phonopy total-IFC path | QE polar-q2r path |
| --- | --- | --- |
| Stored $`\Phi(\mathbf R)`$ | $`\Phi_{\mathrm{VASP}}^{\mathrm{total,SC}}`$: total periodic IFCs represented by the finite supercell | $`\Phi_{\mathrm{q2r}}^{\mathrm{short}}`$: QE has already removed its dipole term |
| Dipole term removed before interpolation | $`\mathrm{IFFT}_{\mathbf q_m}[D_{\mathrm{Gonze}}^{\mathrm{dipole}}]`$: inverse transform from the commensurate q mesh to real space | None: subtracting again would remove the dipole interaction twice |
| Fourier interpolation | $`\mathrm{FT}_{\mathrm{WS}}`$: Gonze/Phonopy Wigner--Seitz transform | $`\mathrm{FT}_{\mathrm{q2r}}`$: native q2r replica and phase convention |
| Dipole term restored at finite q | $`D_{\mathrm{Gonze}}^{\mathrm{dipole}}(\mathbf q)`$ | $`D_{\mathrm{QE}}^{\mathrm{rigid}}(\mathbf q)`$ from QE `rigid.f90` |
| Directional term at exact Gamma | Directional limit is contained in $`D_{\mathrm{Gonze}}^{\mathrm{dipole}}(\mathbf q;\hat{\mathbf q})`$ | $`D_{\mathrm{QE}}^{\mathrm{directional}}(\hat{\mathbf q})`$; $`\delta_{\mathbf q,\Gamma}`$ turns it on only at Gamma |

## Where each term appears in the upstream programs

For stable source-line citations, the links below are pinned to [Phonopy 2.17.1](https://github.com/phonopy/phonopy/tree/b67269df9ff46149550db06ebb652850bae1d1cc) and [Quantum ESPRESSO 7.6](https://github.com/QEF/q-e/tree/9f93ddec427d2b9a45bb72d828c6d324f62fcabd). The Phonopy matrices are calculated at test time from the pinned plain-text inputs; they are not stored outputs attributed to Phonopy 2.17.1. The QE reference tensors and matdyn outputs were generated with QE 7.6. Thus, each symbol in the two equations can be traced to the implementation that defines its numerical convention without overstating the provenance of the fixtures.

| Formula term | Upstream function and lines | Corresponding kALDo PR code | What those lines do |
| --- | --- | --- | --- |
| $`\Phi_{\mathrm{VASP}}^{\mathrm{total,SC}}`$ | Phonopy [`DynamicalMatrixGL.__init__`, lines 550--598](https://github.com/phonopy/phonopy/blob/b67269df9ff46149550db06ebb652850bae1d1cc/phonopy/harmonic/dynamical_matrix.py#L550-L598) | kALDo [`SecondOrder.calculate_nac_short_range_force_constants`, lines 222--275](../blob/bl/qe-gonze-nac-integration/kaldo/observables/secondorder.py#L222-L275) | Accepts the original total finite-supercell IFCs and begins the Gonze short-range construction. |
| $`D_{\mathrm{Gonze}}^{\mathrm{dipole}}(\mathbf q)`$ | Phonopy [`_get_Gonze_dipole_dipole` and `_get_c_recip_dipole_dipole`, lines 763--840](https://github.com/phonopy/phonopy/blob/b67269df9ff46149550db06ebb652850bae1d1cc/phonopy/harmonic/dynamical_matrix.py#L763-L840) | kALDo [`_get_dd_base`, lines 581--654](../blob/bl/qe-gonze-nac-integration/kaldo/controllers/nac.py#L581-L654) and [`dynamical_matrices`, lines 1095--1220](../blob/bl/qe-gonze-nac-integration/kaldo/controllers/nac.py#L1095-L1220) | Converts q and its Gamma direction to Cartesian reciprocal coordinates, evaluates the screened reciprocal dipole tensor, contracts the Born charges, and applies mass weighting. |
| $`\mathrm{IFFT}_{\mathbf q_m}[D_{\mathrm{Gonze}}^{\mathrm{dipole}}(\mathbf q_m)]`$ | Phonopy [`make_Gonze_nac_dataset`, lines 675--726](https://github.com/phonopy/phonopy/blob/b67269df9ff46149550db06ebb652850bae1d1cc/phonopy/harmonic/dynamical_matrix.py#L675-L726), followed by [`DynmatToForceConstants._sum_q`, lines 447--476](https://github.com/phonopy/phonopy/blob/b67269df9ff46149550db06ebb652850bae1d1cc/phonopy/harmonic/dynmat_to_fc.py#L447-L476) | kALDo [`SecondOrder.calculate_nac_short_range_force_constants`, lines 222--275](../blob/bl/qe-gonze-nac-integration/kaldo/observables/secondorder.py#L222-L275) and [`_inverse_transform_dynmats_to_force_constants`, lines 1264--1292](../blob/bl/qe-gonze-nac-integration/kaldo/controllers/nac.py#L1264-L1292) | At every commensurate q, subtracts the dipole matrix from the ordinary dynamical matrix and inverse-transforms the remainder into short-range IFCs. |
| $`\mathrm{FT}_{\mathrm{WS}}[\cdots](\mathbf q)+D_{\mathrm{Gonze}}^{\mathrm{dipole}}(\mathbf q;\hat{\mathbf q})`$ | Phonopy [`_run_py_dynamical_matrix`, lines 316--357](https://github.com/phonopy/phonopy/blob/b67269df9ff46149550db06ebb652850bae1d1cc/phonopy/harmonic/dynamical_matrix.py#L316-L357), and [`DynamicalMatrixGL._compute_dynamical_matrix`, lines 749--761](https://github.com/phonopy/phonopy/blob/b67269df9ff46149550db06ebb652850bae1d1cc/phonopy/harmonic/dynamical_matrix.py#L749-L761) | kALDo [`_short_range_dynamical_matrix_many`, lines 837--882](../blob/bl/qe-gonze-nac-integration/kaldo/controllers/nac.py#L837-L882) and [`dynamical_matrices`, lines 1095--1220](../blob/bl/qe-gonze-nac-integration/kaldo/controllers/nac.py#L1095-L1220) | Phase-averages the shortest Wigner--Seitz vectors, Fourier-transforms the short IFCs, and restores the matching Gonze dipole matrix. |
| $`\Phi_{\mathrm{q2r}}^{\mathrm{short}}`$ | QE [`do_q2r`, lines 257--268](https://github.com/QEF/q-e/blob/9f93ddec427d2b9a45bb72d828c6d324f62fcabd/PHonon/PH/do_q2r.f90#L257-L268), then [lines 291--336](https://github.com/QEF/q-e/blob/9f93ddec427d2b9a45bb72d828c6d324f62fcabd/PHonon/PH/do_q2r.f90#L291-L336) | kALDo [`read_second_order_qe_matrix`, lines 29--130](../blob/bl/qe-gonze-nac-integration/kaldo/interfaces/qe_io.py#L29-L130) and [`read_q2r_header`, lines 382--535](../blob/bl/qe-gonze-nac-integration/kaldo/interfaces/qe_io.py#L382-L535) | Reads the original IFC body while preserving the q2r flag, lattice gauge, q grid, Ewald alpha, dielectric tensor, and Born charges that prove the body is already dipole-subtracted. |
| $`\mathrm{FT}_{\mathrm{q2r}}[\Phi_{\mathrm{q2r}}^{\mathrm{short}}](\mathbf q)`$ | QE [`matdyn::frc_blk`, lines 1134--1232](https://github.com/QEF/q-e/blob/9f93ddec427d2b9a45bb72d828c6d324f62fcabd/PHonon/PH/matdyn.f90#L1134-L1232) | kALDo [`_dynamical_matrix_from_second_order`, lines 46--76](../blob/bl/qe-gonze-nac-integration/kaldo/observables/secondorder.py#L46-L76), selected in [`dynamical_matrices`, lines 1095--1143](../blob/bl/qe-gonze-nac-integration/kaldo/controllers/nac.py#L1095-L1143) | Reconstructs the q2r short-range body with kALDo's established native replica Fourier gauge; it does not reinterpret that body as Gonze Wigner--Seitz IFCs. |
| $`D_{\mathrm{QE}}^{\mathrm{rigid}}(\mathbf q)`$ | QE [`rigid::rgd_blk`, lines 15--45](https://github.com/QEF/q-e/blob/9f93ddec427d2b9a45bb72d828c6d324f62fcabd/PHonon/PH/rigid.f90#L15-L45) and [lines 99--246](https://github.com/QEF/q-e/blob/9f93ddec427d2b9a45bb72d828c6d324f62fcabd/PHonon/PH/rigid.f90#L99-L246), restored by [`matdyn::setupmat`, lines 1268--1281](https://github.com/QEF/q-e/blob/9f93ddec427d2b9a45bb72d828c6d324f62fcabd/PHonon/PH/matdyn.f90#L1268-L1281) | kALDo [`_QERigidIonKernel.rigid_ion_tensor`, lines 353--380](../blob/bl/qe-gonze-nac-integration/kaldo/controllers/nac.py#L353-L380) and [`correction`, lines 411--444](../blob/bl/qe-gonze-nac-integration/kaldo/controllers/nac.py#L411-L444) | Uses QE's fixed Ewald cutoff, onsite compensation, native charge projections, and finite-q $`(\mathbf G+\mathbf q)`$ sum, then converts and mass-weights the restored tensor for kALDo. |
| $`\delta_{\mathbf q,\Gamma}D_{\mathrm{QE}}^{\mathrm{directional}}(\hat{\mathbf q})`$ | QE [`rigid::nonanal`, lines 250--313](https://github.com/QEF/q-e/blob/9f93ddec427d2b9a45bb72d828c6d324f62fcabd/PHonon/PH/rigid.f90#L250-L313), called by [`matdyn`, lines 635--680](https://github.com/QEF/q-e/blob/9f93ddec427d2b9a45bb72d828c6d324f62fcabd/PHonon/PH/matdyn.f90#L635-L680) | kALDo [`_QERigidIonKernel.nonanalytic_tensor`, lines 382--409](../blob/bl/qe-gonze-nac-integration/kaldo/controllers/nac.py#L382-L409), gated by [`correction`, lines 411--444](../blob/bl/qe-gonze-nac-integration/kaldo/controllers/nac.py#L411-L444) | Evaluates the Born-charge numerator divided by $`\hat{\mathbf q}^{\mathsf T}\epsilon_\infty\hat{\mathbf q}`$ and adds it only for the exact-Gamma directional limit. |

The two lines describe the same long-range electrostatics, not two different phonon theories. Their only difference is provenance: VASP/Phonopy supplies total finite-supercell IFCs, while polar q2r supplies an already-short-range IFC body. The invariant is therefore **one dipole subtraction and one matching restoration**. There is no numerical fallback that guesses the convention.

For the generic Gonze path, kALDo applies the Born-charge acoustic sum rule to a private copy, matching Phonopy without changing user-owned data. For q2r, the tensors are preserved exactly because they must reproduce the term that QE removed using its lattice gauge, Ewald parameter, q grid, and fixed `rgd_blk` cutoff. If the Born charges are zero, no dipole term is restored and the ordinary harmonic Fourier path is used.

## What the added tests measure

The tests deliberately compare several levels of the calculation. Define

```math
\Delta D(\mathbf q) = D_{\mathrm{NAC\ on}}(\mathbf q) - D_{\mathrm{NAC\ off}}(\mathbf q).
```

At the lowest level, the tests compare the unmass-weighted force-constant tensor in the native reference units. This is the most sensitive check of the Ewald sum, lattice orientation, Born-charge contraction, atom phases, and Gamma direction. Other tests compare the mass-weighted dynamical matrix used by kALDo, and the final tests compare phonon frequencies and group velocities. Agreement at all three levels is important: a plausible frequency spectrum by itself can hide a wrong tensor convention.

### Numerical reference tests

| Test and number of cases | Reference | Quantity compared | Acceptance criterion | What it establishes |
| --- | --- | --- | --- | --- |
| [QE `rigid.f90` golden test](../blob/bl/qe-gonze-nac-integration/kaldo/tests/test_qe_q2r_nac.py#L65-L93): one deliberately low-symmetry three-atom cell at one finite q and one directional Gamma limit | Arrays emitted by unmodified QE 7.6 `rigid.f90`; the fixture records the exact source SHA-256 | Raw, unmass-weighted $`D_{\mathrm{QE}}^{\mathrm{rigid}}`$ at finite q, and $`D_{\mathrm{QE}}^{\mathrm{rigid}}+D_{\mathrm{QE}}^{\mathrm{directional}}`$ at Gamma, in Ry/bohr$`^2`$ | `rtol = 2e-13`, `atol = 3e-16` | The Python kernel transcribes QE's tensor formula, Ewald convention, anisotropic dielectric response, off-diagonal Born charges, and directional Gamma term to essentially floating-point precision. A separate assertion reconstructs the conversion and $`1/\sqrt{M_iM_j}`$ mass weighting into kALDo units. |
| [QE Bravais campaign](../blob/bl/qe-gonze-nac-integration/kaldo/tests/test_qe_nac_bravais_reference.py#L48-L66): 14 q2r inputs, one for every three-dimensional Bravais class | Transparent QE 7.6 `matdyn.x` NAC-on and NAC-off outputs generated from the checked-in q2r and calculation inputs | $`\Delta D(\Gamma)`$ as a raw, unmass-weighted complex tensor in Ry/bohr$`^2`$, using the reduced direction $`[1,0,0]`$ | Relative Frobenius error $`\lVert D_{\mathrm{kALDo}}-\Delta D_{\mathrm{QE}}\rVert_F/\lVert\Delta D_{\mathrm{QE}}\rVert_F < 5\times10^{-6}`$ | The native q2r lattice, atom, phase, and Cartesian-axis conventions remain correct across all lattice geometries. The `oF` case is a documented synthetic strained-NaCl structure because that Bravais class is absent from the selected PhononDB data. |
| [Gonze/Phonopy campaign](../blob/bl/qe-gonze-nac-integration/kaldo/tests/test_nac_gonze_phonondb26.py#L148-L189): 26 polar materials covering 13 of 14 three-dimensional Bravais classes and 76 matrix comparisons in total | Phonopy calculates NAC-on and NAC-off matrices at test time from 26 pinned, plain-text `phonopy_params.yaml` inputs | The isolated generic-path $`\Delta D(\mathbf q)`$ after the complete mesh sampling $`\rightarrow`$ inverse transform $`\rightarrow`$ short-range interpolation $`\rightarrow`$ dipole restoration pipeline, in kALDo mass-weighted units. Every material uses two fixed finite Cartesian q vectors; 24 cases with a nontrivial supercell also use one exactly commensurate finite q. | Relative Frobenius error $`\le 10^{-5}`$ when $`\lVert\Delta D\rVert_F\ge10^{-8}`$; absolute error $`\le10^{-9}`$ for cancellation cases | The Gonze path matches Phonopy as a complete operation, including non-diagonal supercell matrices, Wigner--Seitz phases, inverse transformation, Born-charge ASR, and arbitrary finite-q restoration. It does not merely compare one internal helper. |
| [Anisotropic wurtzite GaN integration](../blob/bl/qe-gonze-nac-integration/kaldo/tests/test_gan_nac.py#L72-L119): five q-point spectra plus four symmetry/derivative checks | Phonopy frequency spectra for the same force constants and NAC parameters; internal finite differences for velocity | Twelve frequencies at Gamma and four finite q points; Gamma--A transverse degeneracy; Gamma LO--TO splitting; analytic group velocity versus the dispersion slope; and the diagonal heat-flux identity relating $`S_{ij}`$ to velocity | Frequencies within 15 cm$`^{-1}`$; transverse pairs within 0.1 cm$`^{-1}`$; velocity ratio within 1% of $`2\pi`$; flux-derived velocity within $`10^{-3}`$ | A non-cubic, anisotropic polar crystal exercises behavior that cubic NaCl/MgO can hide, including derivative and transport quantities downstream of the dynamical matrix. |
| [Loaded QE MgO end-to-end test](../blob/bl/qe-gonze-nac-integration/kaldo/tests/test_nac_api.py#L206-L229): one real q2r input at reduced q $`[0.3,0,0.3]`$ | Six pinned reference frequencies | The six final mass-weighted harmonic frequencies, in THz | `rtol = atol = 2e-7` | Parsing the q2r provenance, selecting the QE kernel, restoring the long-range tensor, diagonalizing the result, and exposing the public `HarmonicWithQ.frequency` output work together. |

The 26 Phonopy materials and 14 QE Bravais inputs are two independent campaigns, not 26 inputs exercised through both conventions. Their readable fixtures and provenance are documented in the [Phonopy campaign README](../blob/bl/qe-gonze-nac-integration/kaldo/tests/data/input/gonze-phonopy/README.md), [QE Bravais README](../blob/bl/qe-gonze-nac-integration/kaldo/tests/data/input/qe76-bravais-reference/README.md), and [`rigid.f90` generator README](../blob/bl/qe-gonze-nac-integration/kaldo/tests/data/input/qe76-rigid-f90/README.md). The QE Bravais directories retain the SCF, `ph.x`, `q2r.x`, and paired `matdyn.x` inputs as well as the relevant readable outputs; the Phonopy directory retains the decompressed YAML inputs and records each official NIMS download URL and checksum.

### Convention and integration tests

| Test group | Quantity or invariant tested | Failure prevented |
| --- | --- | --- |
| [QE harmonic contracts](../blob/bl/qe-gonze-nac-integration/kaldo/tests/test_qe_nac_harmonic_contract.py#L26-L75) | A loaded polar q2r file owns a `_QERigidIonKernel`, has no Gonze supercell mapping, never requests Gonze short-range subtraction, and rejects an incompatible BvK remeshing request | Double subtraction of the dipole term, accidental generic fallback, and silently changing the q2r sampling convention |
| [q2r header and unit tests](../blob/bl/qe-gonze-nac-integration/kaldo/tests/test_qe_q2r_nac.py#L85-L126) | QE Ewald alpha, q grid, dielectric tensor, Born tensors, masses, cached reciprocal vectors, onsite tensor, and the full native-to-kALDo unit/mass bridge | Losing provenance while parsing, transposing a tensor, silently enforcing the generic Born-charge ASR on QE data, or applying the wrong units |
| [Generic transform and tensor helpers](../blob/bl/qe-gonze-nac-integration/kaldo/tests/test_nac_helpers.py#L154-L217) | Dielectric quadratic form, Cartesian Born contraction, explicit rejection of unknown conventions, and equality of the native IFC forward transform at Gamma and two finite q points | Axis-order errors, accepting ambiguous input provenance, and changing the established force-constant Fourier gauge |
| [NAC option propagation](../blob/bl/qe-gonze-nac-integration/kaldo/tests/test_nac_option_propagation.py#L40-L99) | The identical `nac_bvk_supercell_matrix` reaches heat capacity, two-dimensional heat capacity, population, and QHGK conductivity consumers | Frequencies, eigenvectors, thermodynamic observables, and transport observables being evaluated from inconsistent harmonic matrices |
| [Fixture provenance](../blob/bl/qe-gonze-nac-integration/kaldo/tests/test_nac_gonze_fixture_provenance.py#L9-L30) | All 26 YAML SHA-256 values, source dataset pages, official archive URLs, pinned source index, and absence of an opaque generation script | A scientifically irreproducible or silently changed reference campaign |

The scientific scope of this PR is three-dimensional periodic polar crystals. It does not claim validated two-dimensional LO--TO electrostatics. QE q2r remeshing onto a different Born--von Karman cell is also rejected explicitly rather than approximated by the generic path.


gbarbalinardo (changes requested):
@bkhli A lot of good code. I ran the full suite and rechecked the numbers myself: 286 passed plus 26/26 phonondb. The ASR issue from #262 is fixed (q=0 dipole row sum 7e-12), GaN now agrees with phonopy to 0.01 cm^-1, and I get 1e-4 cm^-1 on the triclinic, monoclinic and trigonal phonondb cases with my own full-dispersion checks. The sha-pinned references and the contract tests are nice, and nac.py is very well documented.

Three blockers.

1. `qe_io.py` is tagged GPL-2.0-only, and `qe_lattice_from_ibrav` is transcribed from QE's `latgen.f90`. kaldo is BSD-3, we can't merge GPL code. Please reimplement the ibrav table independently (phonopy has BSD equivalents) or drop it. If neither works let's talk before merging.

2. NAC velocities use a transposed conversion. `harmonic_with_q.py:221` and `:338` do `primitive_cell.T @ dq_cart`, but with `q_cart = inv(cell) @ q_red` it should be `primitive_cell @ dq_cart`. The `.T` only cancels for symmetric cell matrices, which is why the NaCl tests don't catch it. On GaN the in-plane v_x is off by +95%, on MgO v_x has the wrong sign, and sij uses the same tensors, so kappa of polar non-cubic materials is wrong through RTA and QHGK. This came in with my Jul 11 commit on the #262 branch, sorry. Removing the `.T` gives FD agreement to 1e-4 on all axes, both NAC paths, every q I tested. Please also bring back the gradient-vs-dispersion test from #299 on a non-symmetric cell, the z-only GaN slope test misses exactly this.

3. The QE path doesn't match matdyn.x off Gamma. The polar `qe_q2r` branch uses the plain replica transform, so MgO at q=(0.3,0,0.3) gives [237.55, 240.26, 366.35, 423.82, 424.83, 583.89] vs matdyn [239.76 x2, 367.69, 422.93 x2, 582.65]: up to 2.2 cm^-1 off, and the TA/TO degeneracies split by 2.7 and 1.0 cm^-1. `test_nac_api.py:212` pins these values at rtol 2e-7, so the test locks in the wrong numbers. Your own WS machinery already fixes it: rebuilding the same body with it reproduces matdyn to 0.004 cm^-1 with exact degeneracies. Please switch the polar QE path to the WS weighting and pin the matdyn values with a degeneracy assertion, like the test #299 had.

Smaller, please also address:

4. There's no NAC off switch, and `Phonons(**kwargs)` silently swallows `is_nac=False` or any typo'd NAC kwarg. Please reject unknown kwargs and document how to disable NAC for a polar file.
5. The 26-case suite does `importorskip("phonopy")` but phonopy isn't in requirements or the CI image, so the validation can skip without anyone noticing. Please add it as a test extra and make CI fail if the suite skipped.
6. `test_nac_velocity.py` still has zero test functions. Please delete it or populate it.
7. The 14 `without-nac.fc` fixtures (~12 MB) can be generated with your own `strip_q2r_nac` and no test reads them. Please generate them at test time.
8. `test_gan_nac.py` still has `atol=15.0` with a "worst ~13 cm^-1" comment; the worst I measure is 0.0101. Please tighten and fix the comment.
9. Unused imports in `harmonic_with_q.py` (`Grid`, `json`, `math`, `Path`, `ase.io`, the `nac` alias) and the empty section banners in `secondorder.py`. Please clean up.
10. Please add a release note on the numeric changes for polar materials, including the MgO pinned values moving and `is_unfolding` being a no-op on the NAC path.

Since this contains all of #262, once it lands we close #262 as superseded, and Nik's commits ride along in the history.

bkhli:
All the comments should be addressed in the additional commits.

This PR also fixed several bugs in is_unfolding code path. The unit test also covers more about the consistent behavior between is_unfolding=True with phonopy. 

The feature of interactions between distance_threshold, is_unfolding, is_nac will be addressed in the future. 

Release note draft:

> Polar-material results may change after this release. Corrected QE interpolation restores agreement with QE 7.6 `matdyn.x` (up to 2.2 cm^-1 in the bundled MgO reference), while corrected NAC group velocities change some mode norms by up to approximately 43% and now agree with finite-difference dispersions. These changes may affect RTA, full-BTE, and QHGK thermal conductivities, so users should recompute polar-material results. NAC is detected automatically by default; use `is_nac=False` for an explicit NAC-off calculation. When NAC is active, `is_unfolding` does not change the result because the required Wigner--Seitz interpolation is already applied internally.

PR #304, opened 2026-08-18
Fix origin-dependent IFC interpolation and refactor of Grid

bkhli:
This PR built on top of #303. #303 gave a full quality exmaine on IFC2. This gives a full quality examin on IFC3 and fixing several bugs. 

Changes:

1. Remove `is_unfolding` flag as the name is not obvious. It changes to `ifc_interpolation="<method>"`. 
2. A total refactor on Grid class. 
3. More accurate end-to-end testing.

PR #305, opened 2026-08-18
IFC interpolation (clean rebase of #304)

gbarbalinardo:
Clean rebase of #304 (which contains #303) onto current main, as 7 logical commits. All commits are authored by @bkhli; I only reverted cosmetic churn (reformatting, requoting, docstring-only rewrites) back to main's text where AST equivalence is provable, moved the design doc into `docs/docsource/ifc_interpolation_design.md`, and dropped unread fixture logs. Source diff shrinks from +7448/-2507 to +6476/-1965.

Outputs are bit-identical to #304's head: frequencies, velocities, and RTA kappa on the Si qe-sheng and GaN NAC fixtures differ by exactly 0.0. Subset suites pass with only the known macOS arm64 tolerance failures, same set as #304.

Supersedes #303 and #304.

Remaining before undraft:

- [x] cubic isotropy residual on qe-sheng Si under `auto` (xx off up to 43% at 3x3x3, no decay with grid)
- [x] `use_q_symmetry=True` raises for expanded IFC3 support (sheng/eskm formats); needs support or a warned fallback
- [x] harmonic assembly ~4.5x slower, ~n^2 loop; vectorize (amorphous path first)
- [x] charges-without-dielectric raise breaks nonpolar inputs with plain charge arrays; gate on Born-shaped or nonzero charges and mention `is_nac=False`
- [x] calibrate the six platform-sensitive test tolerances
- [x] release note with the quantified rebaselines (qe-sheng Si RTA -81.6%, vasp-sheng -21.7%, ge d3q +3.75%, TDEP -0.06%, eskm unchanged)

Update: all six items are done, in 7 follow-up commits.

The isotropy residual was the fixture, not the code. `si-crystal/qe/FORCE_CONSTANTS_3RD` was written in the standard diamond basis while POSCAR and espresso.ifc2 use the x-mirrored one. Mirroring the file restores isotropy to 0.04% and moves the qe-sheng RTA pin from 0.83 to 14.86 W/mK at test settings, so the old -81.6% rebaseline number is superseded; the corrected table is in the design doc. The retained ShengBTE reference keeps byte-exact copies of the inputs it consumed, and the rate test runs on those.

Beyond the checklist: `is_unfolding` gets a one-release deprecation shim instead of a hard error, the IFC3 interpolation cache is keyed on content instead of `id()`, and the harmonic assembly rewrite is bit-identical on the amorphous path (0.28 s to 0.02 s per assembly on the 216-atom fixture).

Full suite on macOS arm64: 456 passed, 1 skipped (the deferred IFC4 case), plus all 26 phonondb NAC cases.

Update: restructured as a stacked series for review: #306 (interpolation core fix, existing tests only), #307 (NAC engine, Nik's and Bohan's commits), #308 (validation suites and fixtures). The stack's final tree additionally reconciles with current main (#283) and is fully green (522 passed) on macOS arm64. This PR stays as the reference for the original single-branch rebase.


gbarbalinardo:
Superseded by the stacked series #306 -> #307 -> #308, which also reconciles with current main. Keeping the branch as reference.

Issue #261, opened 2026-05-04
Gonze-Lee NAC correction


niklundgren:
I'd like to get the Gonze-Lee NAC correction working for more than one type of input and use the two other terms (real-space and limiting error term) so that we are maximally capable when dealing with materials with a non-zero dielectric 

It should also be better documented in our theory section.

PR #308, opened 2026-08-19
Validation suites and fixtures for interpolation and NAC

gbarbalinardo:
Last PR of the stack (base: the NAC rebuild). All new validation suites, fixtures, and docs; existing files receive appended tests only.

- Interpolation validation (@bkhli): origin-invariance suites across every loader, the amorphous 32-atom regenerable fixture, TDEP non-diagonal references, ShengBTE offset preservation, and the pinned external ShengBTE transport oracle with byte-exact retained inputs.
- NAC validation (@niklundgren, @bkhli): the GaN/NaCl suites, 26 pinned phonopy references (phonopy is optional: the suite importorskips, no new dependencies), 14 QE 7.6 Bravais references, and provenance contracts.
- Rebaselines, arm64 tolerance calibration with per-site measurements, and the interpolation design doc.

The stack was validated end to end against the full merge of #305 with current main; the review surface on pre-existing files stays minimal and the public API stays the historical `is_unfolding` boolean.

Stack: #306 (core fix), #307 (NAC), then this one.


bkhli:
Some issues: 
1. There is a merge conflict here before to resolve before it can be merged. 
2. A cache issue did not catch in previous PR: `SecondOrder.get_nac_short_range_force_constants()` only verify with supercell matrix, but not IFC2, born effective charges etc. If these parameters changed, the cache are still reused https://github.com/nanotheorygroup/kaldo/blob/c1bd1c2152a427a368f782a5fd6fae8f13b8cfa7/kaldo/observables/secondorder.py#L83-L90
3. In last PR remaining issue, we exposed Phonon class can have `nac_q_direction`, so that there should have a unit test on it and the doc should also be updated as well. 
4. In current CI pipeline, `requirements.txt` does not have phonopy. The Docker image does not have contain this library. So currently, all tests in `test_nac_gonze_phonondb26.py` are skipped in the CI. Are we going to enable this or not in CI?
5. `docs/docsource/ifc_interpolation_design.md` should we keep this doc in the repo? maybe part of #303 's theory explanation with IFC interpolation should go into the dev doc, which can also resolves #261. 
6. `test_nac_helpers.py` has a lot functions not used now.

gbarbalinardo:
@bkhli thanks, in order:

1. merged main in (7a1503b9), no history rewrite. github shows mergeable now. the only conflict was test_nac_qe.py: this branch deletes it in favour of test_nac_api.py, and #307 had added tests to it in the meantime. resolved by keeping the deletion and carrying those tests over, see 3.
2. agreed, and since #307 is on main the fix goes there: #317 keys the short-range cache on a digest of the IFCs, Born charges, dielectric and supercell. same PR fixes the two other cache holes found reviewing #307: ifc_cache_key did not change with the engine or with nac_q_direction, and Conductivity never set _nac_requested so nac_off and nac_on kappa shared a folder.
3. the forwarding test #307 added (test_phonons_forwards_nac_q_direction) had been lost in the earlier merge of this branch, which is the gap you saw. d193138c restores it together with the guard and dispersion tests, in test_nac_api.py, and api_phonons.rst now documents nac_q_direction on Phonons (it still said the grid was pinned to (1,0,0)).
4. cd274349 adds pip install phonopy to the build-and-test job. pure python, no compiled deps, and 26 pinned references skipped in CI are worth nothing.
5. kept in the Developer Guide where it already was, review-order section and PR-relative wording dropped (133d239b) so it reads as the theory reference #261 asks for. if you would rather see the physics sections in a separate theory page, say so and I split it.
6. nine unused helpers and their imports removed (8be3a105).

gbarbalinardo:
CI on the merged branch: 545 passed, 31 skipped (build 3998). the 26 phonopy references in test_nac_gonze_phonondb26.py all ran and passed; the remaining skips are the pre-existing cumulant and non-diagonal ones.

bkhli (approved):
Looks good now. We merge either order of this or #317. After one merge, the other needs to rebase/merge.

Issue #302, opened 2026-08-07
Parallelize cumulant F1/F2 q-loop and Phase-5 sampling (n_workers)


gbarbalinardo:
Follow-up to #283, from the how-do-I-use-more-threads question.

Why one core today: BLAS auto-threads large matrix ops, but the two hot spots are Python loops over small einsums, which BLAS cannot help: (1) the (q1, q2) double loop in cumulant/free_energy.py (F1/F2), and (2) the per-configuration loop in cumulant/thermodynamics.py Phase 5.

Design (kaldo conventions, same as the anharmonic projection):

* Add n_workers to F1_from_fc / F2_from_fc / F1_vectorized / F2_vectorized. Chunk the q1 list (already IBZ-reduced) across workers via kaldo.parallel.get_executor; each worker gets the flattened tables + eigenvector table (a few MB, picklable) and returns partial (F, S, Cv) sums; reduce in q1 order so results stay deterministic. Test: n_workers=2 equals serial to 1e-12 on the vendored fixtures.
* Add n_workers to cumulant_thermo Phase 5. Shard nconf across workers; per-worker RNG via numpy SeedSequence(seed).spawn(n_workers) (document that the stream differs from serial); per-worker LAMMPS instance built from lammps_cmds (strings pickle fine), or a pickled ASE calculator on the calculator= path; concatenate V/V2/V3/V4/V2_tilde shards in worker order.
* Guard BLAS oversubscription: set OMP_NUM_THREADS=1 in workers (or document it), otherwise n_workers x BLAS threads thrashes.

This also unblocks the F0-in-CI wish from #283: with n_workers the end-to-end ASE test can afford enough configs to pin F0 statistically.

ejmeitz:
I believe I have the calculation of `V4` (fourth order potential energy from taylor series) parallelized, but nothing else is.

PR #310, opened 2026-08-24
Parallelize cumulant F1/F2, harmonic, and Energy Sampling

ejmeitz:
Parallelizes the calculation of F1/F2 over the IBZ, parallelizes potential energy sampling by spawning multiple LAMMPS calculator instances, parallelizes harmonic Brillouin zone sum.

Runtime is down to ~30 minutes from 8.5 hours. Julia code is ~5 minutes still. 

The calculation of the potential energy (Phase 5) still seems to be a big bottleneck. About 600s here compared to 120s in Julia.

## Summary by CodeRabbit

* **New Features**
  * Added configurable process-based parallelism for harmonic calculations, free-energy evaluations, and Phase-5 sampling.
  * Added the `n_workers` option to cumulant thermodynamics and related free-energy APIs.
  * Improved adaptive quantum smearing using Brillouin-zone geometry and analytical group velocities.
  * Added parallel support for calculator-based and sampling workflows.

* **Bug Fixes**
  * Improved handling of empty interaction data and thermodynamic table formatting.
  * Updated quantum reference values for affected Neon calculations.

* **Tests**
  * Added regression coverage comparing parallel and serial results and validating worker-count errors.

### DIFF
--- a/kaldo/cumulant/_phase5_worker.py
+++ b/kaldo/cumulant/_phase5_worker.py
@@ -1,0 +1,144 @@
+"""Phase-5 config worker (one LAMMPS per process).
+
+This module is imported by spawn workers. It must not import
+``kaldo.cumulant.thermodynamics`` or ``kaldo.forceconstants`` itself; those
+pull TensorFlow. Parent-package ``__init__`` may still run on spawn — that
+is a separate issue. Keep *this* file's imports limited to sampler,
+contractors, the energy evaluator, and ASE.
+
+Each worker owns:
+  * its own ``LAMMPSlib`` / ASE calculator + ``BatchEnergyEvaluator``
+  * a jumped RNG stream on a pickled ``SCSampler``
+  * in-process V2/V3/V4 (no inner thread pool)
+"""
+from __future__ import annotations
+
+import os
+
+import numpy as np
+
+from kaldo.cumulant._energy_evaluator import BatchEnergyEvaluator
+
+
+def _lammps_log_path(spec=None):
+    """Per-process LAMMPS log so n_workers>1 does not collide on one file."""
+    log_dir = None
+    if spec:
+        log_dir = spec.get("log_dir")
+    if not log_dir:
+        log_dir = os.environ.get("KALDO_CUMULANT_LAMMPS_LOG_DIR", "/tmp")
+    os.makedirs(log_dir, exist_ok=True)
+    return os.path.join(log_dir, f"cumulant_thermo_lammps_{os.getpid()}.log")
+
+
+def _load_contractors(spec):
+    """Unpickle a tiny contractor table, or rebuild from ``tdep_folder``.
+
+    Production Phase-5 shards must not pickle ``phi4`` into every worker
+    (hundreds of MB per process). Pass ``tdep_folder`` and rebuild locally.
+    Tests may still ship a small ``contractors`` object.
+    """
+    contractors = spec.get("contractors")
+    if contractors is None:
+        folder = spec.get("tdep_folder")
+        if not folder:
+            raise ValueError("Phase-5 worker needs contractors= or tdep_folder=")
+        from .contractors import SCContractors
+        contractors = SCContractors.from_tdep_folder(
+            folder, include_fourth=spec.get("include_fourth", True),
+        )
+    return contractors
+
+
+def _instantiate_calculator(calculator):
+    """Return a calculator instance from a class, factory, or live instance."""
+    if calculator is None:
+        return None
+    if isinstance(calculator, type):
+        return calculator()
+    # ASE calculators have ``calculate``; factories are bare callables.
+    if callable(calculator) and not hasattr(calculator, "calculate"):
+        return calculator()
+    return calculator
+
+
+def build_energy_eval(spec):
+    """Build a worker-local ``BatchEnergyEvaluator`` from picklable spec fields."""
+    from ase import Atoms
+
+    atoms = Atoms(
+        spec["species"],
+        positions=spec["eq"],
+        cell=spec["cell"],
+        pbc=True,
+    )
+    if spec.get("lammps_cmds") is not None:
+        from ase.calculators.lammpslib import LAMMPSlib
+
+        kwargs = dict(spec.get("lammps_kwargs") or {})
+        atoms.calc = LAMMPSlib(
+            lmpcmds=list(spec["lammps_cmds"]),
+            keep_alive=True,
+            log_file=_lammps_log_path(spec),
+            **kwargs,
+        )
+    else:
+        atoms.calc = _instantiate_calculator(spec["calculator"])
+        if atoms.calc is None:
+            raise ValueError("Phase-5 worker needs lammps_cmds= or calculator=")
+    return BatchEnergyEvaluator(atoms, spec["eq"])
+
+
+def run_phase5_loop(nconf, sampler, contractors, energy_eval, verbose=False, logger=None):
+    """Serial Phase-5 loop over ``nconf`` draws. Used by n_workers=1 and by each worker."""
+    V = np.zeros(nconf)
+    V2 = np.zeros(nconf)
+    V3 = np.zeros(nconf)
+    V4 = np.zeros(nconf)
+    V2_tilde = np.zeros(nconf)
+    dV2_tilde_dT = np.zeros(nconf)
+    import time
+    t0 = time.time()
+    for n in range(nconf):
+        u, z = sampler.draw_with_z()
+        V[n] = energy_eval.energy(u)
+        V2[n] = contractors.V2(u)
+        V3[n] = contractors.V3(u)
+        V4[n] = contractors.V4(u)
+        V2_tilde[n], dV2_tilde_dT[n] = sampler.V2_tilde_and_dT_from_z(z)
+        if verbose and logger is not None and (n + 1) % max(1, nconf // 10) == 0:
+            logger.info(f"  n={n+1}/{nconf}  ({time.time()-t0:.1f}s)")
+    return V, V2, V3, V4, V2_tilde, dV2_tilde_dT
+
+
+def eval_u_chunk(spec):
+    """Evaluate V/V2/V3/V4 on a fixed list of displacements (no RNG).
+
+    Used by tests so parallel vs serial can match bit-for-bit on contractors
+    and energy, independent of jumped sample streams.
+    """
+    contractors = _load_contractors(spec)
+    energy_eval = build_energy_eval(spec)
+    us = spec["us"]
+    n = len(us)
+    V = np.empty(n)
+    V2 = np.empty(n)
+    V3 = np.empty(n)
+    V4 = np.empty(n)
+    for i, u in enumerate(us):
+        V[i] = energy_eval.energy(u)
+        V2[i] = contractors.V2(u)
+        V3[i] = contractors.V3(u)
+        V4[i] = contractors.V4(u)
+    return V, V2, V3, V4
+
+
+def run_phase5_chunk(spec):
+    """Process-pool entry: jumped RNG, own LAMMPS, serial contractors."""
+    contractors = _load_contractors(spec)
+    sampler = spec["sampler"]
+    sampler.rng = np.random.Generator(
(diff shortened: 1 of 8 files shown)


ejmeitz:
#302

kaldo/cumulant/thermodynamics.py
```diff
@@ -73,13 +74,55 @@ class CumulantResult:
...
+    from ._harmonic_worker import freq_q_chunk, init_freq_worker
+
+    q_points = np.asarray(ph._reciprocal_grid.unitary_grid(is_wrapping=False))
+    n_q = int(q_points.shape[0])
+    n_cpu = n_workers if n_workers is not None else (os.cpu_count() or 1)
+    n_cpu = max(1, min(int(n_cpu), n_q))
+    chunks = [c for c in np.array_split(q_points, n_cpu) if c.size]
+    second = ph.forceconstants.second
+    # Materialize IFC2 as numpy so spawn pickles arrays, not TF tensors.
+    second._dynmat = np.asarray(second.dynmat)
```
gbarbalinardo, line 92: this overwrites the cached tf tensor on the caller's object. after it runs, `second.dynmat.numpy()` crashes downstream (elastic_prop, unfolding, distance_threshold) and test fixtures stay mutated. keep the numpy copy local and pass it via initargs, don't write it back on `second`.

kaldo/cumulant/thermodynamics.py
```diff
@@ -310,65 +374,132 @@ def cumulant_thermo(
...
-        V2[n] = contractors.V2(u)
-        V3[n] = contractors.V3(u)
-        V4[n] = contractors.V4(u)
-        V2_tilde[n], dV2_tilde_dT[n] = sampler.V2_tilde_and_dT_from_z(z)
-        if verbose and (n + 1) % max(1, nconf // 10) == 0:
-            logging.info(f"  n={n+1}/{nconf}  ({time.time()-t0:.1f}s)")
+    if not is_parallel(n_workers):
+        at = Atoms(species_sc, positions=sc_pos_eq_A, cell=sc_cell_A, pbc=True)
+        if calculator is not None:
+            at.calc = calculator
```
gbarbalinardo, line 413: the one i care most about: we now have a third calculator convention. a factory works at n_workers=2 and crashes here at 1 (this branch assigns the callable without instantiating). validate_parallel_calculator runs only at phase 5, after the expensive phases, and the calculator isn't touched before then. and `_instantiate_calculator`'s `callable and not hasattr("calculate")` sniff differs from the finite difference rule.

can we validate once, early, next to `_check_n_workers`:
```python
if is_parallel(n_workers):
    validate_parallel_calculator(calculator, "cumulant_thermo")
```
and use the fd rule in both the serial branch and the worker:
```python
calc = calculator() if callable(calculator) else calculator
```
then `_instantiate_calculator` goes away. one contract for the whole repo. (validate_parallel_calculator lets lambdas through and they die at submit with a raw PicklingError, that's a hole in the helper, i'll harden it separately)

kaldo/cumulant/_phase5_worker.py
```diff
@@ -0,0 +1,163 @@
...
+    return V, V2, V3, V4
+
+
+def run_phase5_chunk(spec):
+    """Process-pool entry: jumped RNG, own LAMMPS, serial contractors."""
+    contractors = _load_contractors(spec)
+    sampler = spec["sampler"]
+    sampler.rng = np.random.Generator(
+        np.random.PCG64(spec["seed"]).jumped(int(spec["worker_id"]))
+    )
```
gbarbalinardo, line 146: `jumped(0)` is the identity, worker 0 draws the same stream as a serial run at the same seed. pooling a serial and a parallel run duplicates a whole worker's samples and biases the bootstrap SE low.
```suggestion
    sampler.rng = np.random.Generator(
        np.random.PCG64(spec["seed"]).jumped(int(spec["worker_id"]) + 1)
    )
```
the comment in test_parallel_phase5.py:170 says the streams are not bit-identical, same bug, fix it too

kaldo/cumulant/thermodynamics.py
```diff
@@ -310,65 +374,132 @@ def cumulant_thermo(
...
+                    completed = int(progress[:, 0].sum())
+                    if verbose and (completed >= next_report or not pending):
+                        logging.info(
+                            f"  Phase 5: {completed}/{nconf} "
+                            f"({100.0*completed/nconf:.0f}%)  "
+                            f"elapsed={elapsed:.1f}s"
+                        )
+                        while next_report <= completed:
+                            next_report += report_every
+        finally:
```
gbarbalinardo, line 494: if one worker raises (or ctrl-c), shutdown waits for every other worker to finish its whole chunk before the error surfaces. with a big nconf that's hours. catch and `executor.shutdown(wait=False, cancel_futures=True)`, same for the f1/f2 pools

kaldo/cumulant/thermodynamics.py
```diff
@@ -310,65 +374,132 @@ def cumulant_thermo(
...
+        n_cpu = n_workers if n_workers is not None else (os.cpu_count() or 1)
+        n_cpu = max(1, min(int(n_cpu), nconf))
+        counts = [len(c) for c in np.array_split(np.arange(nconf), n_cpu)]
+        counts = [c for c in counts if c > 0]
+        if verbose:
+            logging.info(f"Phase 5: starting {len(counts)} processes "
+                         f"({nconf} configs, one LAMMPS each) ...")
+        from multiprocessing import shared_memory
+        progress_shape = (len(counts), 8)  # one cache line per worker
+        progress_shm = shared_memory.SharedMemory(
```
gbarbalinardo, line 443: nconf=0 with workers reaches `SharedMemory(size=0)` and dies with an opaque ValueError, after phases 1-4 already ran. check nconf early and send nconf<2 to the serial branch

kaldo/cumulant/free_energy.py
```diff
@@ -132,8 +135,370 @@ def build_psi4_realspace_v(QD, q1_cart, q2_cart):
...
+            except FileNotFoundError:
+                pass
+            except Exception:
+                pass
+        self._shms.clear()
+
+
+def _init_q1_worker(thread_cap, spec, scalars):
+    from kaldo.parallel.executor import _init_worker_thread_caps
+    _init_worker_thread_caps(int(thread_cap))
```
gbarbalinardo, line 270: `_init_worker_thread_caps` sets env vars, but in a spawned worker numpy is already imported by the time the initializer body runs (it comes in while unpickling initargs), so blas keeps its default thread count and the pools oversubscribe (8 workers x 8 blas threads, parallel can lose to serial). `threadpoolctl.threadpool_limits(thread_cap)` in the initializer binds after import. i can harden the helper upstream so all callers get it.

kaldo/cumulant/contractors.py
```diff
@@ -156,85 +153,29 @@ def V2(self, u_flat):
...
-        return float(sum(f.result() for f in futs))
+        """V_4 in eV."""
+        u1 = u_flat[self.a1_4]
+        u2 = u_flat[self.a2_4]
+        u3 = u_flat[self.a3_4]
+        u4 = u_flat[self.a4_4]
+        n = u1.shape[0]
+        if n == 0:
+            return 0.0
+        t = np.matmul(self._phi4_27_3, u4[:, :, None])[..., 0]
```
gbarbalinardo, line 175: stacked matmul doesn't thread the batch dimension, so serial phase 5 lost its multi-core kernel: the old threaded split is ~2x faster at n=2e5 (60ms vs 30ms at 4 threads), and OMP_NUM_THREADS / KALDO_CONTRACT_THREADS became no-ops. keep the threaded kernel for n_workers=1, cap it to 1 thread inside pool workers

kaldo/cumulant/free_energy.py
```diff
@@ -318,6 +677,35 @@ def build_psi3_realspace(TD, q2_cart, q3_cart):
...
+    e1_vectors = _triplet_mode_vectors(np.conj(e1), TD["a1"])
+    return np.einsum("nka,nabc->nkbc", e1_vectors, TD["ifc"], optimize=True)
+
+
+def _build_psi3_modes_triplet_batch(TD, left_q1, e2, e3, q2_cart, q3_cart):
+    """Build mode-space Psi3 for a batch without a real-space scatter."""
+    e2_vectors = _triplet_mode_vectors_batch(np.conj(e2), TD["a2"])
+    e3_vectors = _triplet_mode_vectors_batch(np.conj(e3), TD["a3"])
+    phase = np.exp(-1j * (q2_cart @ TD["lv2c"].T + q3_cart @ TD["lv3c"].T))
+    middle = np.einsum("nkbc,qnlb->qnklc", left_q1, e2_vectors, optimize=True)
```
gbarbalinardo, line 705: batch sizes are hardcoded (`_F1_Q2_BATCH_SIZE=4`, `_F2_Q2_BATCH_SIZE=8`) and the intermediates scale as n_terms * nb^2 complex128: at 1e5 triplets, nb=24 this einsum output is ~22 GB per process. inputs the old serial loop handled can now OOM. derive the batch from a byte budget, `max(1, budget // (n_terms * nb * nb * 16))`

kaldo/cumulant/_phase5_worker.py
```diff
@@ -0,0 +1,163 @@
+"""Phase-5 config worker (one LAMMPS per process).
+
+This module is imported by spawn workers. It must not import
+``kaldo.cumulant.thermodynamics`` or ``kaldo.forceconstants`` itself; those
+pull TensorFlow. Parent-package ``__init__`` may still run on spawn — that
```
gbarbalinardo, line 5: on this: every spawned worker still imports tensorflow through `kaldo/cumulant/__init__` (free_energy and this file are tf-free). four pools x 8 workers is 20-60s of startup and multi-GB transient rss. pep 562 in the package `__init__` fixes it:
```python
def __getattr__(name):
    if name in ("cumulant_thermo", "CumulantResult", "print_thermo_table"):
        from kaldo.cumulant import thermodynamics
        return getattr(thermodynamics, name)
    raise AttributeError(name)
```

gbarbalinardo (changes requested):
ethan, good stuff in here: kaldo's executor instead of a new pool, jumped rng streams, workers rebuilding contractors instead of pickling phi tables, bit-exact tests at 1e-12. right skeleton.

one ask before the rest: the F2 smearing default change comes out of this pr. the default moved from tdep's smearingparameter to adaptive_sigma and the Ne reference numbers were re-pinned in the same pr (U2 moves -5.3%). maybe adaptive is the better default, but nobody bisecting a physics change will look in a parallelization pr. make it its own pr and we discuss it there. this one stays bit-compatible with main on the serial path.

the bugs are inline. short version:
- `second._dynmat` gets overwritten with numpy, later `.numpy()` calls crash
- calculator handling: factory works at n_workers=2 and crashes at 1, validation runs after the expensive phases, `_instantiate_calculator` redoes what validate_parallel_calculator already does
- worker 0 rng is the serial stream, pooled runs duplicate samples
- one failed worker blocks until all the others finish
- blas caps in the initializer don't bind under spawn
- v3/v4 lost the threaded kernel, serial ~2x slower per config
- the batched kernels can OOM inputs the old loop handled
- nconf=0 with workers crashes after phases 1-4

follow-ups, open issues instead if you prefer: shared store for phi/evecs (right now 16 workers each re-parse the tdep tables), dispatch_with_resume for the n_workers plumbing (the copies already diverge), one pool across the four phases.

nits: print_thermo_table output format changed on a public function, the pid lammps logs in /tmp never get cleaned and `spec["log_dir"]` has no producer, two docstrings use em dashes.

the bones are right, the rest is mechanical :slightly_smiling_face:

ejmeitz:
Thanks. I've rebased this on top of a new PR which changes how I compute the sigma value.

I made most of the changes. Some I did now are below:

- Inner ThreadPool for V3/V4 stays out. `n_workers` is the only parallelism control. There are already a lot of knobs to be turned here. I would prefer only one way of parallelizing. 
- On the F1/F2 batch sizes: 4 and 8 were the fastest in the microbenchmarks, not a RAM choice. I added kwargs for these as they were just hard coded before.

ejmeitz (commented):
Done with review. Trying to make the message go away so it will run CI....

PR #312, opened 2026-08-26
Default F2 quantum smearing to adaptive_sigma instead of smearingpara…

ejmeitz:
I swapped this back because this is what my Julia code uses. It makes a minor difference in the final sig-figs, but with this I get agreement to more digits. It causes some disagreement for smaller mesh sizes but I'd rather have a 1-1 copy for now.

## Summary by CodeRabbit

- **Improvements**
  - Improved quantum free-energy calculations with geometry-aware default smearing and analytic group velocities.
  - Applied consistent smearing results across the full wave-vector grid.
  - Improved force-constant reconstruction reliability by validating replica information and deriving translation dimensions from available data.

- **Tests**
  - Updated Neon quantum reference values for free energy, internal energy, entropy, and heat capacity.


gbarbalinardo (approved):
approving: LDT's free_energy.jl uses _adaptive_sigma with the same prefactor and clamps, and the julia pins in the regression harness were taken under adaptive sigma, so this restores the validated default. thanks for splitting it out.

one ask before merge: the sigma block is calculate_broadening_tdep from kaldo.controllers.anharmonic (bit-identical, same radius, same clamps). call it directly and delete smearingparameter, adaptive_sigma (the shims) and calculate_smearingparameter_tdep, all callerless now.

two optional: with atoms=None fall back to the identity mapping over the full grid instead of raising, since only the IBZ shortcut needs atoms; and one line in the description on where the new Ne numbers come from.

on coderabbit: the 2π thread is wrong (TDEP pairs the no-2π radius with dω/dq_phys and puts the 2π in the prefactor, same as here); the other two are pre-existing, leave them.

PR #316, opened 2026-09-10
fix isotopic g factor lookup when NIST is unreachable

afiorent:
compute_gfactor caught only urllib.error.HTTPError, so the legacy database took over only when a server answered with an error status. The cases that actually occur on a compute node - no DNS, no route, a blackholed connection - raise URLError or TimeoutError and crashed the calculation instead.

Catch the whole failure surface, narrow the try to the download alone, and pass a timeout so a firewall that drops packets cannot hang the run. A downloaded table that does not cover every element in the structure now falls back too, rather than raising KeyError partway through the loop.

physics.nist.gov also sits behind Cloudflare now, whose default bot ruleset bans the 'Python-urllib' agent signature with a 403 (error 1010). ase's download_isotope_data cannot override that header, so issue the request here under a user agent that names kaldo; the body still goes through ase's parse_isotope_data, so NIST format changes keep tracking upstream. The bundled legacy values were checked against the live table and agree exactly for all 84 elements with a natural isotopic composition.

Elements with no stable composition previously produced a silent nan - null in the legacy database, zero abundances from NIST - which propagated into the bandwidth. They now raise a ValueError naming the element and pointing at the g_factor argument of Phonons.

## Summary by CodeRabbit

* **New Features**
  * Isotopic g-factor data is retrieved from the NIST database when available, with bundled-data fallback for offline use or missing coverage.
  * Elements without natural isotopic composition now produce clear validation errors.
  * Added configurable non-analytic correction (NAC) support, including supercell and q-direction settings, with validation for incompatible configurations.

* **Tests**
  * Added coverage for isotope downloads, fallback behavior, validation, and NAC-related functionality.


gbarbalinardo:
Thanks @afiorent. Please address coderabbit feedback.

gbarbalinardo:
Please remove Claude as co-author.

